PR #208, opened 2021-02-04
Michaeljeffrey/docker local testing

michaeldjeffrey:
I'm iffy about if we want to compile the LoRaMac-Node stuff when we build the image, but it cuts down on testing time.


macpie:
Going to keep this as a local change

PR #226, opened 2021-02-26
Keep router from falling on its face when hotspots report region_subregion for as923

michaeldjeffrey:
- Hotspots are planing on reporting their subregion as part of AS923, if
they provide, we should respect that.

- Move maybe fetching locations back to every offer. If router is
redeployed and a device does not try to join again, we won't know what
region it's in.


michaeldjeffrey:
Formatting issues fixed @macpie

macpie (approved):

PR #240, opened 2021-03-11
Improve trace logging

macpie:
This is to improve device log tracing by providing device id when possible to avoid having duplicate messages from other device sharing same `devaddr`

### DIFF
--- a/src/device/router_device_routing.erl
+++ b/src/device/router_device_routing.erl
@@ -375,10 +375,7 @@ maybe_buy_join_offer(Offer, _Pid, Device) ->
                     PHash = blockchain_state_channel_offer_v1:packet_hash(Offer),
                     case bloom:set(BFRef, PHash) of
                         true ->
-                            case maybe_multi_buy(Offer, 10, Device) of
-                                ok -> ok;
-                                {error, _} = Error -> Error
-                            end;
+                            maybe_multi_buy(Offer, 10, Device);
                         false ->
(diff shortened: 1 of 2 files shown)


src/device/router_device_routing.erl
```diff
@@ -344,7 +375,10 @@ maybe_buy_join_offer(Offer, _Pid, Device) ->
                     PHash = blockchain_state_channel_offer_v1:packet_hash(Offer),
                     case bloom:set(BFRef, PHash) of
                         true ->
-                            maybe_multi_buy(Offer, 10, Device);
+                            case maybe_multi_buy(Offer, 10, Device) of
```
michaeldjeffrey, line 378: This case seems unnecessary

src/device/router_device_routing.erl
```diff
@@ -931,12 +969,16 @@ get_chain() ->
             Chain
     end.
 
--spec handle_offer_metrics(any(), ok | {error, any()}, non_neg_integer()) -> ok.
-handle_offer_metrics(#routing_information_pb{data = {eui, _}}, ok, Time) ->
+-spec handle_offer_metrics(
+    any(),
```
michaeldjeffrey, line 973:
```suggestion
    #router_information_pb{},
```

michaeldjeffrey (changes requested):

michaeldjeffrey (approved):

PR #253, opened 2021-03-23
2021.03.19.0 Patch

macpie:
- Fix downlink info reported to Console
- Let packet pass when fcnt is > current even if delivery is late
- Increase some eunit timeouts for local runs
- Fix event_uplink crash when payload is `undefined`
- Deliver devaddr in hex to console
- Fix metric for ws state on websocket_terminate
- Deliver packet info on downlink queued
- Fix: sometimes replay packets might be mistaken as late packets\
- Fix potential duplicate join requests https://github.com/helium/router/issues/255
- Fix validate frame replay window https://github.com/helium/router/issues/256


src/router_utils.erl
```diff
@@ -266,13 +272,14 @@ event_downlink(
             {_, true} -> {downlink_confirmed, <<"Confirmed data down sent">>};
             {_, false} -> {downlink_unconfirmed, <<"Unconfirmed data down sent">>}
         end,
+    Payload = blockchain_helium_packet_v1:payload(Packet),
     Map = #{
         id => router_utils:uuid_v4(),
         category => downlink,
         sub_category => SubCategory,
         description => Desc,
         reported_at => erlang:system_time(millisecond),
```
macpie, line 281: comment here

michaeldjeffrey (approved):

Issue #257, opened 2021-03-29
CN470 tx power too high


lthiery:
Transmit no higher than 17

![image](https://user-images.githubusercontent.com/1606640/112871152-d0950d00-9073-11eb-8f41-013a0e2b7ad0.png)

Is this handled for Europe? @macpie

macpie:
I am guessing this depends on region?

lthiery:
Yes, transmit power will depend on region. As does datarate for example, so I would try to keep them compartmentalized in similar ways.

lthiery:
Found this fella this morning @macpie @michaeldjeffrey 
![image](https://user-images.githubusercontent.com/1606640/112880795-c6790b80-907f-11eb-9913-ee313c5b6727.png)

macpie:
This had been merged

PR #258, opened 2021-03-30
Break out downlink strength per region

michaeldjeffrey:
- see #257


JayKickliter:
For the most part (1), these values should roughly correspond to the first entry in the [uplink power table](https://github.com/helium/router/blob/master/src/lora/lorawan_mac_region.erl#L553) for the respective region.

1: there are exceptions because LoRaWAN

macpie (approved):

PR #260, opened 2021-03-30
Macpie/router http channel url check

macpie:


michaeldjeffrey (approved):

.env-template
```diff
@@ -21,5 +23,8 @@ ROUTER_CONSOLE_SECRET=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 # Max time to wait for uplinks in ms
 ROUTER_FRAME_TIMEOUT=300
 
-# Turn on/off xor filter worker (anything else than true is off)
-ROUTER_XOR_FILTER_WORKER=true
\ No newline at end of file
+# Turn on/off xor filter worker (anything else than true is off, disabled by default)
+ROUTER_XOR_FILTER_WORKER=true
+
+# Turn on/off HTTP integration url checks like DNS, IP... (anything else than false is on, enabled by default)
```
mfalkvidd, line 29: Looks to me like this is disabled by default. Comment says enabled by default.

Is it safe to set this setting to false? If it is safe, why does the option to check exist at all?

macpie: This is just the template but you are right it is confusing, I ll update template to be `true` instead

macpie: See https://github.com/helium/router/pull/261

disk91:
This should be reported in the console/.env-template file as the router is usually get as an image and the env file comes from the console environement when following the documentation.

Issue #270, opened 2021-04-02
Can not disable ADR


JayKickliter:
Once a device has ADR enabled, it's not possible to disable it. I've tried both of the following with no success:

- removing an ADR-enabled label from a device
- toggling ADR off on a label while leaving that label attached to a device

macpie:
Fixed

PR #271, opened 2021-04-02
Default ADR to false when doing a get_device

macpie:
Fix for https://github.com/helium/router/issues/270


JayKickliter (approved):
I confirm this fixes both scenarios #270

PR #283, opened 2021-04-12
Add extra metrics

macpie:


michaeldjeffrey (approved):
I did see a couple times where we'd get blocks before the api. So that might be a negative number sometimes.

macpie:
Negative is ok, we will just alert over a predefined threshold

PR #284, opened 2021-04-13
add util scripts for saving logs and monitoring head block

michaeldjeffrey:


scripts/monitor_blocks.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
macpie, line 1: We got some stats for this now

michaeldjeffrey: Can we leave it in?
I've been using it on prod to make sure router is catching back up (metrics haven't been deployed yet).
And it could be useful for open source users who don't hook hook up metrics.

macpie: Ok seems fair

macpie (approved):

PR #285, opened 2021-04-13
Charging org depending on if offer was received or not

macpie:


michaeldjeffrey (approved):
Do we need any logging for packets that come without offers?

PR #294, opened 2021-04-26
Michaeljeffrey/ch9907/downlink queued item should no longer appear

michaeldjeffrey:


src/device/router_device_worker.erl
```diff
@@ -885,6 +885,7 @@ handle_info(
                 DownlinkPacket,
                 Region
             ),
+            ok = maybe_send_queue_update(Device1, State),
```
macpie, line 888: Idk about this one as we do not actually update the queue but rather potentially remove a downlink from it

michaeldjeffrey: When we send a downlink they don't want to see it in the side panel any more.

michaeldjeffrey: Unless it's a confirmed down. But once it's out of the queue, they want that shown on the frontend

macpie: sold

macpie (approved):

PR #298, opened 2021-04-28
Enhance http integration with dynamic query parameters

michaeldjeffrey:
I'm a little dubious about interpolating the keys of query params. But it's easy enough to rip out, thought I'd add it first to have the discussion.


macpie (approved):
Good stuff

PR #304, opened 2021-05-07
Special case EU868 downlink power

michaeldjeffrey:
Regional Params are moving to the blockchain. Or we're going to need to
have a pow-wow of what's a good general answer to not blasting downlinks
out at the max power.


macpie:
We have to update core that old branch is gone. I did it in https://github.com/helium/router/pull/305. Otherwise looks good 👍🏻

PR #308, opened 2021-05-14
Add option to customize frame time out for disco mode

macpie:


michaeldjeffrey (approved):
I wonder if we can use some of this stuff to speed up some tests.

macpie:
> I wonder if we can use some of this stuff to speed up some tests.

We already have an option to change timeout, we would need to bake it in the test also (wait_for_data stuff)

PR #312, opened 2021-05-18
Ignore metrics for disco mode

macpie:


michaeldjeffrey (approved):
Added a comment so future me can remember why.

lgtm

PR #317, opened 2021-05-25
Michaeljeffrey/as923 updates

michaeldjeffrey:
Miners are going to start reporting subregions for AS923 bands. 
Outside of joining, the lora spec doesn't differentiate between the sub-regions.
So we intercept the region and map it to the top level region before continuing. 
This lets the device remain as ignorant as possible that it might be carrying a region with extra information. And we get to keep the subregions for potential debugging help.

#### Still need:
- [x] Frequency Lists for Subregions `3` and `4`


michaeldjeffrey:
Getting rolled into a PR that includes DLSettings updates.

PR #330, opened 2021-06-01
Expose Region default rx2 datarate index

### LINKED ISSUE #327: JoinAccept DLSettings should not be always 0
Dear,

[This line](https://github.com/helium/router/blob/master/src/device/router_device_worker.erl#L1210) always sends `DLSettings = 0` which is an issue at least in US915 and AU915 but maybe in other regions, I did not check (AU923 with dwell time?):

The RX2 datarate is given by the network in the joinAccept frame thanks to the parameter `RX2DataRate`, which is contained in DLSettings:
image

As per LoRaWAN specification (1.0.2 to 1.0.4) US915 downlinks are on DR8 to DR13, as Halium network server is providing a value of `RX2DataRate = 0` the device will never receive any downlinks on RX2 datarate.

Note: Some stacks allows the network server to use `RX2DataRate = 0` but downlinks will never work, some others will refuse the joinAccept

michaeldjeffrey:
Fixes #327

Ref: DLSettings from LLS 1.0.4.pdf
https://lora-alliance.org/resource_hub/lorawan-104-specification-package/

### DIFF
--- a/src/device/router_device_worker.erl
+++ b/src/device/router_device_worker.erl
@@ -1207,7 +1207,8 @@ handle_join(
 
 -spec craft_join_reply(atom(), binary(), binary(), binary()) -> binary().
 craft_join_reply(Region, AppNonce, DevAddr, AppKey) ->
-    DLSettings = 0,
+    DR = lorawan_mac_region:window2_dr(Region),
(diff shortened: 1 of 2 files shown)


michaeldjeffrey:
Getting rolled into a PR that includes AS923 updates.

PR #341, opened 2021-06-16
Michaeljeffrey/sc updates

michaeldjeffrey:


src/router_sc_worker.erl
```diff
@@ -206,23 +206,28 @@ schedule_next_tick() ->
...
-            {ok, Id0} = open_next_state_channel(State),
-            State#state{in_flight = [Id0 | NewF], tombstones = []};
-        Other ->
-            %% don't do anything
-            lager:info("active_count = ~p, standing by...", [Other]),
-            State#state{in_flight = [], tombstones = []}
+maybe_start_state_channel(#state{in_flight = InFlight, tombstones = Tombstones, chain=Chain} = State) ->
+    Opened = get_opened_count(),
+    Active = get_active_count(),
+    {ok, Limit} = blockchain:config(max_open_sc, blockchain:ledger(Chain)),
```
macpie, line 212: Maybe grab that on init

macpie: on post init

michaeldjeffrey: Good thinking

rebar.config
```diff
@@ -20,7 +20,9 @@
     {httpc_aws, {git, "https://github.com/helium/httpc-aws.git", {branch, "master"}}},
     {websocket_client, "1.3.0"},
     {erlang_v8, {git, "https://github.com/Vagabond/erlang_v8.git", {branch, "master"}}},
-    {blockchain, {git, "https://github.com/helium/blockchain-core.git", {branch, "master"}}},
+    {blockchain,
```
macpie, line 23: dont forget to run `rebar3 upgrade blockchain` to upgrade the deps in lock file

src/router_sc_worker.erl
```diff
@@ -112,18 +102,29 @@ handle_info(post_init, #state{chain = undefined} = State) ->
             erlang:send_after(500, self(), post_init),
             {noreply, State};
         Chain ->
+            Limit = mac_sc_open(Chain),
```
vihu, line 105:
```suggestion
            Limit = max_sc_open(Chain),
```

src/router_sc_worker.erl
```diff
@@ -201,57 +203,51 @@ terminate(_Reason, _State) ->
 %% Helper funs
 %% ------------------------------------------------------------------
 
+-spec mac_sc_open(Chain :: blockchain:blockchain()) -> pos_integer().
```
vihu, line 206:
```suggestion
-spec max_sc_open(Chain :: blockchain:blockchain()) -> pos_integer().
```

src/router_sc_worker.erl
```diff
@@ -201,57 +203,51 @@ terminate(_Reason, _State) ->
...
+mac_sc_open(Chain) ->
```
vihu, line 207:
```suggestion
max_sc_open(Chain) ->
```

src/router_sc_worker.erl
```diff
@@ -326,13 +318,23 @@ get_nonce(PubkeyBin, Ledger) ->
...
+                lists:sort(
+                    fun(SCIDA, SCIDB) ->
+                        {ActiveSCA, _} = maps:get(SCIDA, SCs),
+                        {ActiveSCB, _} = maps:get(SCIDB, SCs),
+                        blockchain_state_channel_v1:expire_at_block(ActiveSCA) <
+                            blockchain_state_channel_v1:expire_at_block(ActiveSCB)
+                    end,
+                    ActiveSCIDs
+                ),
+            {SoonestSCToExpire, _} = maps:get(SoonestSCIDToExpire, SCs),
```
vihu, line 336: This will blow up if `SoonestSCIDToExpire` is not found in `SCs`, presumably that's the behavior you want here?

macpie: yeah this would be very bad if we had a active SC ID but not the actual SC, this is why I always use `maps:get/2`

test/router_sc_worker_SUITE.erl
```diff
@@ -0,0 +1,189 @@
...
+    meck:expect(blockchain_state_channels_server, state_channels, fun() ->
+        #{sc1 => sc1}
+    end),
+    meck:expect(blockchain_state_channels_server, get_active_sc_count, fun() -> 1 end),
+
+    % Force tick another tick
+    router_sc_worker ! '__router_sc_tick',
+
+    % By then last inflight txn should have cleared and we should still be at 1
+    test_utils:wait_until(fun() ->
```
michaeldjeffrey, line 94: This won't fail. It will retry until it gives up, then carry on.

michaeldjeffrey:
```suggestion
    ok = test_utils:wait_until(fun() ->
```

macpie: I am gonna nuke that test I started with it but the full test should cover all that anyway

src/router_sc_worker.erl
```diff
@@ -326,13 +333,23 @@ get_nonce(PubkeyBin, Ledger) ->
 
 -spec active_sc_expiration() -> {error, no_active_sc} | {ok, pos_integer()}.
 active_sc_expiration() ->
-    case blockchain_state_channels_server:active_sc_id() of
-        undefined ->
+    case blockchain_state_channels_server:active_sc_ids() of
```
michaeldjeffrey, line 336: Can we take a min of all the expirations here, and not have to deal with the potential of the maps blowing up?
```
active_sc_expiration() ->
    case blockchain_state_channels_server:active_scs() of
        [] ->
            {error, no_active_sc};
        ActiveSCs ->
            Expires = [blockchain_state_channel_v1:expire_at_block(SC) || SC <- ActiveSCs],
            {ok, lists:min(Expires)}
    end.
```

PR #342, opened 2021-06-17
Offer cache auto cleanup

macpie:
Offer cache will now clean itself with new offers to avoid weird build ups

Time was also taken down from 10min to 2min


michaeldjeffrey (approved):
If you feel like changing `TenMin` to `ExpireTime` or something I'd be down. But it makes sense without it. 
Or log the difference in size if we care about it.

PR #346, opened 2021-06-22
Move crafting reply closer to sending join

michaeldjeffrey:
newer lora spec supports an optional channel mask in the cflist for
us915 devices. older specs say devices should ignore non-empty cflists,
but real world testing proved this to be a lie.

We are going to alternate filling the cflist with channel mask
information to give each device a chance to join without having to
register what version devices they are in console. They will still get
channel mask updates in the first packet after a successful join. This
will allow newer devices to be a little smarter.

Count join attempts in device worker. Allows for the alternating
cflists, also something we can potentially look at to see who's having
trouble joining.

### DIFF
--- a/test/router_SUITE.erl
+++ b/test/router_SUITE.erl
@@ -11,6 +11,7 @@
     dupes_test/1,
     dupes2_test/1,
     join_test/1,
+    us915_join_cf_list_test/1,
     adr_test/1
 ]).
 
@@ -41,6 +42,7 @@ all() ->
     [
         dupes_test,
         join_test,
+        us915_join_cf_list_test,
         adr_test
     ].
 
@@ -726,7 +728,7 @@ join_test(Config) ->
         AppKey,
         DevNonce
     ),
-    ?assertEqual(CFList, lorawan_mac_region:mk_join_accept_cf_list('US915')),
+    ?assertEqual(CFList, lorawan_mac_region:mk_join_accept_cf_list('US915', _FirstJoinAttempt = 0)),
     %% Check that device is in cache now
     {ok, DB, [_, CF]} = router_db:get(),
     WorkerID = router_devices_sup:id(?CONSOLE_DEVICE_ID),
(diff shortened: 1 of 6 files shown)


Vagabond:
You need to send the CFList correction *after* join if you don't send it on the join packet. I'm not sure this PR does that yet?

test/router_SUITE.erl
```diff
@@ -782,6 +784,70 @@ join_test(Config) ->
...
+
+    {_NetID2, _DevAddr2, _DLSettings2, _RxDelay2, _NwkSKey2, _AppSKey2, CFList2} = test_utils:wait_for_join_resp(
+        PubKeyBin,
+        AppKey,
+        DevNonce2
+    ),
+    %% This CFList is empty. We don't know if the device is using lora version
+    %% older than 1.0.3 and might be choking on the cflist, so we'll send an
+    %% empty one and see if they can join with that. The device will still
+    %% receive a channel mask update in the first packets downlink after join.
```
Vagabond, line 832: I would recommend you test this here to be sure?

Vagabond:
or do we ALWAYS send the follow up cflist as well, as before? That way in case the device ignores the join cflist it will still get corrected?

test/router_SUITE.erl
michaeldjeffrey: Added

michaeldjeffrey:
For US915 always send a follow up cflist, that has remained unchanged.
(added in https://github.com/helium/router/pull/158/)

test/router_SUITE.erl
Vagabond: Well, I mean I was proposing we test we get the cflist in a downlink stiill unless this is already well tested elsewhere

michaeldjeffrey: The link_adr_req command tested in some of the adr tests.
I beefed up this test to make sure the link_adr_req is turning on the same channels as the join cf_list.

Vagabond (approved):

macpie (approved):

PR #348, opened 2021-06-24
Allow device queue to be cleared from downlink payload

michaeldjeffrey:


macpie:
@michaeldjeffrey Can we get some docs about this?

macpie (approved):

PR #354, opened 2021-06-30
Upgrade core to 2021.06.30.0

michaeldjeffrey:


macpie:
@michaeldjeffrey is this needed and if so when?

michaeldjeffrey:
This contains the fix for the peerbook that was thrashing some memory. 

Not immediately necessary, but preparing for tomorrow.

macpie (approved):

PR #360, opened 2021-07-06
Ability to toggle alternating cflist in us915 join

michaeldjeffrey:


src/router_utils.erl
```diff
@@ -608,6 +609,17 @@ join_timeout() ->
...
 
+-spec join_toggle_cflist() -> boolean().
+join_toggle_cflist() ->
+    case application:get_env(router, join_toggle_cflist, true) of
+        [] ->
+            true;
+        Str when is_list(Str) ->
+            erlang:list_to_atom(Str);
+        I ->
+            I
```
macpie, line 620: I think we should force stuff like here https://github.com/helium/router/blob/master/src/device/router_device_devaddr.erl#L333

macpie (approved):

PR #373, opened 2021-07-14
Macpie/misc

macpie:


src/router_utils.erl
```diff
@@ -565,6 +565,12 @@ trace(DeviceID) ->
         ],
         debug
     ),
+    _ = erlang:spawn(fun() ->
+        Timeout = application:get_env(router, device_trace_timeout, 240),
```
michaeldjeffrey, line 569: Will anything happen if this is a string?

src/router_utils.erl
```diff
@@ -533,7 +533,7 @@ lager_md(Device) ->
 trace(DeviceID) ->
     BinFileName = trace_file(DeviceID),
     {ok, Device} = router_device_cache:get(DeviceID),
-    FileName = erlang:binary_to_list(BinFileName) ++ ".log",
+    FileName = "traces/" ++ erlang:binary_to_list(BinFileName) ++ ".log",
```
michaeldjeffrey, line 536: 🙏

src/router_utils.erl
macpie: this is meant to be change manually it is not exposed

michaeldjeffrey (approved):

PR #377, opened 2021-07-19
Fix open_next_state_channel

macpie:


macpie:
Just thought about this but I need to make sure that we do not go over the max time limit

src/router_sc_worker.erl
```diff
@@ -298,18 +298,25 @@ open_next_state_channel(#state{pubkey = PubKey, sig_fun = SigFun, oui = OUI, cha
...
                 %% We set the next SC expiration to the difference between
                 %% current chain height and active plus the expiration_interval
-                abs(ActiveSCExpiration - ChainHeight) + get_sc_expiration_interval()
+                Max = blockchain_utils:approx_blocks_in_week(Ledger),
+                Expiration = abs(ActiveSCExpiration - ChainHeight) + get_sc_expiration_interval(),
+                case Expiration > Max of
+                    false ->
+                        Expiration;
+                    true ->
+                        Expiration - (Expiration - Max)
```
michaeldjeffrey, line 315: Is this `min(Expiration, Max)`? 

Do we want to log if we're changing an expected the expiration, so we can know when we're overreaching?

michaeldjeffrey (approved):

PR #394, opened 2021-08-07
Potential crash when sending events to console

### LINKED ISSUE #379: Frame counter management causing "Uplink dropped: Late" packets over AU915 - Staging
1. Are you using Helium hosted Console or running your own open source Console?
Hosted.

2. If you are on Helium hosted Console, which url are you currently on?
https://staging-console.helium.wtf. 
I observed it for the first time on the production console, but it's still on the staging console.

3. If you are experiencing bugs in Helium hosted Console, please specify your Organization name and Device ID (ex. 3483050b-284d-47e2-8eda-e087fc03a753).
Organization name: My Organization
Device ID: 8c0d27fe-a9ba-41da-b8c3-ce05174d8c06

Please describe your issue:

This issue is happening for devices working with AU915 and the subband 2 channels.

I'll split this description into two main facts, maybe they're intertwined and the solution to the first one could solve the second one.

**1. Join accepts are only logged on the console events when they come from some specific frequencies/channels, such as 917.0 MHz (channel 9), and 917.5 MHz (channel 65).** 
For example, the following image shows a sequence of multiple resets. The device generates the join request and receives the join accept with all of them, however, just the one with 917.0 MHz is recorded by the console.
Capture01

**2.  "Uplink dropped: Late" related to frame counters.**

- From the device side, the frame counter starts from zero for every new join request cycle, for instance when a reset is performed. However, it seems the console/router keeps the last successful uplink logged as a parameter to determine if there is a valid packet even though after a new join request. So, uplinks are dropped ("Uplink dropped: Late") until the frame counter reaches the last successful uplink logged.
Capture02

- Testing with the frame counter from the device side.
Capture03

michaeldjeffrey:
When asking for a channel we assume all frequencies in events going to
console are uplink frequencies. Most of the  time this is okay because
uplinks and downlinks follow more or less the same line. That's not the
case for all regions.

If we can't get a channel assuming an uplink frequency, try assuming
it's a downlink frequency, otherwise we have a good failure on our
hands.

fixes #379

### DIFF
--- a/src/device/router_device_channels_worker.erl
+++ b/src/device/router_device_channels_worker.erl
@@ -735,7 +735,7 @@ format_hotspot(
         snr => blockchain_helium_packet_v1:snr(Packet),
         spreading => erlang:list_to_binary(blockchain_helium_packet_v1:datarate(Packet)),
         frequency => Freq,
-        channel => lorawan_mac_region:f2uch(Region, Freq),
+        channel => lorawan_mac_region:freq_to_chan(Region, Freq),
         lat => Lat,
         long => Long,
         hold_time => HoldTime
--- a/src/lora/lorawan_mac_region.erl
+++ b/src/lora/lorawan_mac_region.erl
@@ -31,7 +31,7 @@
 
 -export([downlink_signal_strength/2]).
 -export([dr_to_down/3]).
--export([window2_dr/1, top_level_region/1, f2uch/2]).
+-export([window2_dr/1, top_level_region/1, freq_to_chan/2]).
 -export([mk_join_accept_cf_list/2]).
 
 -include("lorawan_db.hrl").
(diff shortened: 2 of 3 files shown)


macpie (approved):

johansmacias:
Is this already deployed? Because I'm still facing the same behavior I described in #379. Btw, thanks for putting your hands on this.

michaeldjeffrey:
@johansmacias It has not been deployed yet. We're getting the next release ready. It will be out tomorrow.

zbockaj:
Should this fix apply to other regions also? Because I just got the same problem [#379 ](https://github.com/helium/router/issues/379) with EU868, got "late packets" after rejoin until the frame counter on device incremented enough. Tested on console.helium.com, organization: Protoinit, device_id: 5422571d-5844-4a90-867f-52346becbb70

Issue #396, opened 2021-08-10
Multi packet issues


BFGNeil:
Running the Wio Terminal mapper and I have all packets turned on.

Lots are returning late, even tho its by MS, and often means I get a lost response (even tho one went through)

example here:

<img width="1028" alt="Screenshot 2021-08-09 at 15 25 35" src="https://user-images.githubusercontent.com/79230094/128835991-b9a37ffc-eb93-43d7-8d67-2b767e0b8fcf.png">

jdgemm:
We're investigating.

macpie:
@BFGNeil Could you find out what was the hold time for the initial and late packet? The `hold_time` is a value given to the integrations, you can also use the debug mode to find it.

BFGNeil:
All I can see is: 

{
  "id": "3504a146-a708-4686-94e8-847c6f976060",
  "router_uuid": "1c466a1b-2cd3-4ac1-a23a-ebfce68f672f",
  "category": "uplink_dropped",
  "sub_category": "uplink_dropped_late",
  "description": "Late packet",
  "reported_at": "1628931730065"
}

am I missing something? debug on console doesn't seem to have hold_time

BFGNeil:
"hold_time": 475, late packet does not list a hold_time

macpie:
Looks like the hotspot held on to the packet a little longer than the the threshold (400ms currently). A few fixes were implemented to improve hold time but it does not seems to always be enough.

I believe we should rethink how we timeout for packets and maybe allow longer if we do not intend to send a downlink to the device. We then will be able to capture more packets and avoid the infamous "late packet issue".

This will be my top priority.

macpie:
This should be improved with #410. 

Following our release schedule we will tag today for a week of testing and release next week.

macpie:
This is not fully fixed just yet

macpie:
Moving here https://github.com/helium/router/issues/412

PR #410, opened 2021-08-18
Macpie/ch10173/multi packet issues lots of late packets

macpie:


macpie:
Fix for https://github.com/helium/router/issues/396

michaeldjeffrey (approved):

slavendam:
I did testing on staging. It is much better then before. Before I rarely could get 2 hotspots. :) 
image

I see few potential issues now:
1. There are still Dropped packages, but less then before. Will increase of wait time help to solve this?
2. I see more of double packets. Packets from some hotspots are late for 3-10s, and are sent as new Uplinks.  Even if someone didn't buy Multiple packets, he will get this and will need to pay. These are known as Repeated packets, and should be thrown away according to LoRaWAN specs (security). (see photo - Uplink 19)
3. Sometimes there are 12-15 hotspots, sometimes just 3-4. Device is at same position and sending each 2 minutes. Is it Console problem, or hotspots are busy with something so they miss packet? (see photo - Uplinks 20 and 21 for few, and then 22 for many)

image


image

macpie:
It seems like you are down linking (or acking) on every packet. The improvement here was only for uplink with no downlink in return. We will discuss a way to improve this also.

Issue #424, opened 2021-09-02
Prevent unlimited queueable downlinks


michaeldjeffrey:
Prevent people from adding large amounts of downlinks to a device. 
A device with a large queue can cause certain parts of router to start to slow down. 
And Console would start to have a hard time allowing users to manage their downlinks queues as those requests aren't paginated.

PR #425, opened 2021-09-02
Add queue size limit to device

macpie:
Fix for #424


michaeldjeffrey (approved):
We need to remember to put the env vars we've changed recently into the `.env-router` in Console

PR #434, opened 2021-09-16
microsoft azure iot hub pre built integration

michaeldjeffrey:
When I was testing on Dev there was still about a 30% failure rate with the mqtt connection. I haven't been able to track down what's causing it yet.


src/channels/router_azure_channel.erl
```diff
@@ -0,0 +1,349 @@
...
+    lager:warning("rcvd unknown call msg: ~p", [_Msg]),
+    {ok, ok, State}.
+
+handle_info({publish, PublishPayload}, #state{azure = Azure, channel = Channel} = State) ->
+    case router_azure_connection:mqtt_response(Azure, PublishPayload) of
+        {ok, Payload} ->
+            Controller = router_channel:controller(Channel),
+            router_device_channels_worker:handle_downlink(Controller, Payload, Channel);
+        {error, _} ->
+            ok
```
Vagabond, line 100: Log the error here?

michaeldjeffrey: yes!

src/channels/router_azure_connection.erl
```diff
@@ -0,0 +1,415 @@
...
+    parse_connection_string/1,
+    connection_information/2,
+    generate_http_sas_token/1,
+    generate_mqtt_sas_token/1
+]).
+
+-export_type([azure/0]).
+
+%% longest expiration time for azure is 365 days
+-define(EXPIRATION_TIME, timer:hours(25)).
```
macpie, line 39: Should we generate for longer then?

macpie (approved):

PR #440, opened 2021-09-28
Don't crash the xor worker when it can't get information

michaeldjeffrey:


dpezely (approved):
Seems straight-forward enough...

macpie (approved):

PR #441, opened 2021-09-30
Use blockchain-core's e2qc

JayKickliter:
Upstream e2qc appears to be abandoned. We're already using a forked version in blockchain-core; let's use that.


src/router.app.src
```diff
@@ -17,7 +17,6 @@
         helium_proto,
         observer_cli,
         hackney,
-        e2qc,
```
michaeldjeffrey, line 20: I think this needs to stay in the app deps even if we're not pulling it in directly from rebar.

dpezely (approved):
Confirmed: 

This builds with `make docker-build` (existing erlang:22 debian based image), and `make docker-test` passes.

Builds on `erlang:23-alpine` with accommodations for other dependencies such as patched `inert`.

src/router.app.src
dpezely: it builds and tests fine without re-adding this line.

Output of `rebar3 tree` shows that e2qc is found via another dependency chain, so that might be confirmation enough.

```
└─ router─0.1.0 (project app)
   ├─ blockchain─0.1.0 (git repo)
   │  ├─ e2qc─1.2.1 (git repo)
```

JayKickliter: Honestly, I've never understood the purpose of this list. I'll re-add it if necessary.

PR #446, opened 2021-10-05
Consider channel_correction in calculating frame timeout

michaeldjeffrey:
When channel correction is to be part of a downlink, our frame timeout
needs to be the short window so we have a chance of getting the downlink
to the device within the RX window.

It would be nicer to ask the frame if there are any FOpts going down,
but we don't know that when we're calculating the timeout. It could be
moved to before hand, but that will take some different work.

### DIFF
--- a/test/router_SUITE.erl
+++ b/test/router_SUITE.erl
@@ -1162,7 +1162,7 @@ us915_link_adr_req_timing_test(Config) ->
     %% Sending the third packet [fcnt: 2], just a basic packet. Nothing is going
     %% down the device, so we use the uplink_integration_req event to do our
     %% frame timeout calculation. By this time the device has recognized that
-    %% it's channels haven been corrected, and it should use a longer frame
+    %% its channels haven been corrected, and it should use a longer frame
     %% timeout to increase our chances of multi-buy packets.
 
     Stream !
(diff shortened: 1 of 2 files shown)


dpezely (approved):
Much of the flow remains opaque to me, but of the pieces familiar to me: 
Looks good to me!

test/router_SUITE.erl
```diff
@@ -937,6 +939,311 @@ us915_join_disabled_cf_list_test(Config) ->
...
+        "No downlink to be sent, adr not acknowledged, still short window"
+    ),
+
+    ok = test_utils:ignore_messages(),
+
+    %% ========================================================================
+    %% Sending the third packet [fcnt: 2], just a basic packet. Nothing is going
+    %% down the device, so we use the uplink_integration_req event to do our
+    %% frame timeout calculation. By this time the device has recognized that
+    %% it's channels haven been corrected, and it should use a longer frame
```
dpezely, line 1165:
```suggestion
    %% its channels haven been corrected, and it should use a longer frame
```

michaeldjeffrey: Looks like this was already fixed in not a suggestion

macpie (approved):

Issue #469, opened 2021-10-26
use paginated console endpoints for devices_all and organizations_all


jdgemm:
Use these routes to start
/api/router/devices
/api/router/organizations

{
data: [{...}, {...} ...],
after: "resource_id"
}

Use these routes to paginate
/api/router/devices?after={^resource_id}
/api/router/organizations?after={^resource_id}

Last page will return
{
data: []
}

PR #479, opened 2021-10-29
Paginate API calls

macpie:


macpie:
This is a fix for #469

The goal here is to paginate `/device`  and `/organizations` APIs

dpezely (approved):
It would be worthwhile adding a brief comment about *why* we need `get_devices/4` to be recursively accumulating results and the related interaction/behavior with the upstream call.

dpezely (approved):

michaeldjeffrey (approved):

PR #481, opened 2021-11-01
 Backoff on context creation and add sup

macpie:
- We are not retrying to decode anymore
- Context creation is done async with some backoff if need be
- v8 and custom decoder sup are now under same sup to avoid taking down more than they should

### DIFF
--- a/config/sys.config
+++ b/config/sys.config
@@ -85,28 +85,6 @@
         {log_root, "/var/data/log"},
         {crash_log, "crash.log"},
         {handlers, [
-            {lager_console_backend, [
-                {level, warning},
-                {formatter_config, [
-                    "[",
-                    date,
-                    " ",
-                    time,
-                    "] ",
-                    pid,
-                    " [",
-                    severity,
-                    "]",
-                    {device_id, [" [", device_id, "]"], ""},
-                    " [",
-                    {module, ""},
-                    {function, [":", function], ""},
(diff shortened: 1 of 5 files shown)


dpezely (approved):
Without comparing this code to Ferd's instructions for back-off & retry, this looks good.
(I'm assuming that you've followed the same pattern as used elsewhere in the code since you probably wrote most of those earlier instances anyway...)

michaeldjeffrey (approved):
At some point do we want to provide more fleshed out error messages? For example if someone get's `no_context` that might not mean anything to them.

Also running more than 1 v8 vm for when we get lots of data at the same time. We've seen spikes up to 15ms for single function runs, I suspect most of that is waiting for other contexts to finish.

macpie:
> At some point do we want to provide more fleshed out error messages? For example if someone get's `no_context` that might not mean anything to them.
> 
> Also running more than 1 v8 vm for when we get lots of data at the same time. We've seen spikes up to 15ms for single function runs, I suspect most of that is waiting for other contexts to finish.

Could be, but this does not impact downlink so I am not too worried about the 15ms

macpie:
> Without comparing this code to Ferd's instructions for back-off & retry, this looks good. (I'm assuming that you've followed the same pattern as used elsewhere in the code since you probably wrote most of those earlier instances anyway...)

Yes same as before

PR #490, opened 2021-11-02
Sequence diagrams

dpezely:
From top-level `Makefile`, run `make docs` to generate the .svg and .png files.

As noted in the new `docs/README`, this requires [plantuml](https://github.com/plantuml/plantuml/releases), available via `apt-get` and `brew`.

New sequence diagrams: 
- Erlang processes: initialize supervisors, workers and ETS tables
- Evaluating custom JavaScript decoders via V8 engine
- [more soon... Look for .plantuml files under `docs/` subdirectory]

There's always a balance between insufficient and too much documentation.  The idea is to capture high-level interactions that could make the router's systems architecture and source code more accessible to new staff and the community at large.

These diagrams will NOT cover fine-grained details of program flow.  For that, the source code is always authoritative.  However, some nuances may be identified such as where interdependent modules invoke a `gen_server` versus `gen_event`, etc.

Edit: end-to-end flow of a package through router will be part of a separate PR.


Makefile
```diff
@@ -28,6 +35,10 @@ run: | $(grpc_services_directory)
 docker-build:
 	docker build -f Dockerfile-local --force-rm -t quay.io/team-helium/router:local .
 
+docker-docs:
```
macpie, line 38: Will this extract the created file back to local env?

dpezely: Yes, the generated .png and .svg files will appear on the local laptop/workstation file system because `-v` (mount local volume) is given to `docker run` without needing to extract or copy any files.

macpie (approved):

Issue #484, opened 2021-11-02
Devices intermittently failing to submit data to integrations. `no_channel`


jthiller:
**Device ID that this was noticed on: 59eafe96-fa7b-4e9c-992a-01db8675e4e9**

Over the weekend had a device drop its integrations for periods of time. 
Was working fine as a mapper until I noticed it stopped showing up on the map despite the expectation that there should be signal (was boating in San Francisco Bay). Console reported “no_channel” as the integration message. (See Figure below)
<img src="https://user-images.githubusercontent.com/1965053/139773426-706a06d5-ef53-42b4-bb47-f6a40dc81db5.png" width="360">

The mapper device: 
- has two integrations, mappers and cargo
- has unlimited packet purchasing turned on
- briefly regained connection after initially dropping. (Visible in the screenshot below)
- had uplinks received throughout the ‘dead periods’ (seen in console)
<img src="https://user-images.githubusercontent.com/1965053/139773471-d2ad9f35-6c72-4b34-a157-1fd148e9c5d6.png" width="360">

This has also been noticed by another user in the #mappers discord channel. 
https://discord.com/channels/404106811252408320/730801390431240228/904905401466761237
![image](https://user-images.githubusercontent.com/1965053/139773917-b47f29d7-df4e-4edf-8f7d-aa1100a35aef.png)

macpie:
Thanks for the reports!

I have identified 2 issues here:

1. The database use by Console is having issue due to load, @vicmgs  is working on a patch for this that will be deployed tomorrow.
2. When Console API are not returning a `200 OK` Router decide to take it as an empty response which as the consequence of looking like the device is not attached to any integrations. _(It was implemented that way a while back as communication between Router and Console was "perfect", now with load we need to be a little less defensive and retry if we get an error to make sure that integrations are started correctly on Router's side.)_

I will be issuing a PR to fix this asap.

PR #492, opened 2021-11-02
Retry to get channels if it fails

macpie:
Fix for issue #484

If the API is having issue we do not try refetch channels and just return an empty list. We now, will retry to get channels (integrations) if the first API call failed to ensure proper channel startup

### DIFF
--- a/src/device/router_device_channels_worker.erl
+++ b/src/device/router_device_channels_worker.erl
@@ -41,6 +41,7 @@
 ]).
 
 -define(SERVER, ?MODULE).
+-define(REFRESH_CHANNELS, refresh_channels).
 -define(BACKOFF_MIN, timer:seconds(15)).
 -define(BACKOFF_MAX, timer:minutes(5)).
 -define(BACKOFF_INIT,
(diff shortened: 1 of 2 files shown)


dpezely (approved):
Adding a test for coverage, an optional URL param for `router_console_api:get_device_()` could be used for attempting `http://foo.invalid` (always fails DNS lookup per RFC 6761) thus failing-out of the happy path.

src/device/router_device_channels_worker.erl
```diff
@@ -342,40 +343,46 @@ handle_cast(_Msg, State) ->
...
-                    fun(Channel) -> self() ! {start_channel, Channel} end,
-                    maps:values(APIChannels)
-                ),
-                %% Removing old channels left in cache but not in API call
-                remove_old_channels(EventMgrRef, APIChannels, Channels0)
-        end,
-    {noreply, State#state{channels = Channels1}};
+    case router_console_api:get_channels(Device, self()) of
+        {error, _Reason} ->
+            _ = erlang:send_after(timer:seconds(1), self(), ?REFRESH_CHANNELS),
```
michaeldjeffrey, line 351: We might want to have a retry limit. The other day, a channel was misconfigured on the frontend and wasn't able to be converted, so it would fail, kill the worker, and retry when the next packet came through. 
This would catch that, then retry every second forever.

michaeldjeffrey: Nevermind, this will only fail if the request fails. 
And malformed channels will continue in the same pattern as before.

michaeldjeffrey (approved):

PR #496, opened 2021-11-06
Bring back try mechanism for decoders

michaeldjeffrey:
temporary fix for crashed contexts invalidating all other contexts


macpie (approved):

dpezely (approved):
ok.  Relying upon much of this reverting an earlier commit but haven't fully vetted those differences considering the circumstances of this being a hot patch.
That said, I like the additional test.

PR #499, opened 2021-11-08
Skip Decoders when sending join-req to integration

michaeldjeffrey:
Join frames don't have payloads, so there's nothing to decode.

### DIFF
--- a/src/channels/router_channel.erl
+++ b/src/channels/router_channel.erl
@@ -298,7 +298,7 @@ encode_data(Decoder, #{payload := Payload, port := Port} = TemplateArgs, Channel
             )
     end;
 encode_data(_Decoder, TemplateArgs, _Channel) ->
-    lager:info("encoding join [device: ~p] [channel: ~p]", [
+    lager:debug("encoding join [device: ~p] [channel: ~p]", [
         ?MODULE:device_id(_Channel),
         ?MODULE:unique_id(_Channel)
     ]),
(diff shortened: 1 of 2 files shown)


dpezely (dismissed):
We have a lot of repetition within test code, so at some point we should consider generic/prepopulated maps for `wait_for_console_*()` and insert just the override values.

src/channels/router_channel.erl
```diff
@@ -307,7 +296,13 @@ encode_data(Decoder, #{payload := Payload, port := Port} = TemplateArgs, Channel
                     payload => base64:encode(Payload)
                 })
             )
-    end.
+    end;
+encode_data(_Decoder, TemplateArgs, Channel) ->
+    lager:info("missing payload and port [device: ~p]", [?MODULE:device_id(Channel)]),
+    router_channel_utils:maybe_apply_template(
```
macpie, line 302: I wonder if we should even apply template?

mikev (approved):

src/channels/router_channel.erl
michaeldjeffrey: That's a fair point. They still might be trying to grab `payload` or the likes and will fail on that.

src/channels/router_channel.erl
```diff
@@ -307,7 +296,13 @@ encode_data(Decoder, #{payload := Payload, port := Port} = TemplateArgs, Channel
                     payload => base64:encode(Payload)
                 })
             )
-    end.
+    end;
+encode_data(_Decoder, TemplateArgs, _Channel) ->
+    lager:info("encoding join [device: ~p] [channel: ~p]", [
```
macpie, line 301: Lets just take that log line down to debug
```suggestion
    lager:debug("encoding join [device: ~p] [channel: ~p]", [
```

dpezely (approved):

macpie (approved):

PR #500, opened 2021-11-09
Make sure we fail CI if #downlink{} or #channel{} are changed

michaeldjeffrey:
These tests will mostly protect from fields that are added to the record.
All fields exist in these records exist because we use them somewhere.
If a field name is changed the compiler will warn us.
If a field is removed, same complaint.

### DIFF
--- a/src/channels/router_channel.erl
+++ b/src/channels/router_channel.erl
@@ -563,4 +563,33 @@ hash_test() ->
     Hash = crypto:hash(sha256, erlang:term_to_binary(Channel1)),
     ?assertEqual(Hash, hash(Channel0)).
 
+channel_record_update_test() ->
+    %% This test exists to make sure upgrades are considered if anything needs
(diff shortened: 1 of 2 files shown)


dpezely (approved):
Would it also be useful confirming the number of elements in the resulting record?  I don't recall whether `?assertMatch()` macro does partial-match or comprehensive/exhaustive inspection.

michaeldjeffrey:
> Would it also be useful confirming the number of elements in the resulting record? I don't recall whether `?assertMatch()` macro does partial-match or comprehensive/exhaustive inspection.

If the number of fields change, the tuple won't match.
It's kind of like doing
```
{downlink, _Confirmed, _Port, _Payload, _Channel} = #downlink{}.
```
But gives more of an error message than just `badmatch`.

macpie (approved):

Issue #463, opened 2021-10-23
Server not responding to device sending FCtrl.ADR=true, should send LinkADRReq with faster DataRate & FCtrl.ADR=true


buzzware:
I have a third party node which I can only configure its EUIs, OTAA/ABP and cause it to rejoin. It is in the same room, or within 50 meters of the Sensecap and TTS gateways.
It works fine on TTN/TTS, configured as AU915 sideband 2, LoRaWAN 1.0.3. It joins on SF12 then within a few messages it shifts to SF7 via ADR.
However on Helium it joins with SF7 then uplinks in SF12, which wastes energy at an unacceptable level.
Each uplink is normally repeated with a 3 second gap - perhaps a separate issue, that may be caused/made worse by SF12 and so I am focussing on fixing the SF first.
I have configured the profile for the device, with ADR and CF enabled, then caused it to rejoin. This doesn't change the behaviour.
I then tried decoding the Join-Accept packets, and found that TTS was specifying a CFList and Helium was not.
Then I tried looking into the router source, and found the following :

1) the file c_src/compile.sh line 20 contains "for REGION in US915 EU868 AS923 CN470; do" - should AU915 be there ?
2) the file src/lora/lorawan_mac_region.erl contains calls to mk_join_accept_cf_list for most other plans, but not AU915
3) are there other issues for Helium to fully support AU915 side band 2 ?

Please let me know if there is anything I can do to test changes. I don't have Erlang experience, but have deep experience in other more mainstream languages. This is critical for us to use Helium for our startup.

mikev:
@buzzware - Check router_device_worker.erl:2027
We don't consider ADR at all if it's not enabled. 

I'm not sure how Console reconciles multiple labels with conflicting settings. 
You might have it on in 1 place, then applied something else that turns it off.

Can you provide your device-id?  That will allow us to double-check your device profile settings.  It will also allow us to initiate a trace.

In the spreadsheet row 14 says "uplink" but you've linked to a downlink packet.  Can you double-check your packet decodes?

buzzware:
> @buzzware - Check router_device_worker.erl:2027 We don't consider ADR at all if it's not enabled. ADR must be set in the device's profile.
> 
> I'm not sure how Console reconciles multiple labels with conflicting settings. You might have ADR on in 1 place, then applied something else that turns it off.

I've only had one label applied at a time to my devices, and that profile has ADR on.
[...]

PR #502, opened 2021-11-09
#463 - Set adr bit prior to downlink

mikev:


src/device/router_device_worker.erl
```diff
@@ -1621,6 +1621,7 @@ handle_frame_timeout(
                     fcnt = FCntDown,
                     fopts = FOpts1,
                     fport = Port,
+                    adr = Frame#frame.adr,
```
JayKickliter, line 1624: I don't understand how this line relates to the other new comment (`ADR bit normally must be set if MAC ADR command is downlinked`) in the test suite.

mikev: The end-device on the Uplink sets the ADR bit.  On the Downlink Ack we should also set the ADR bit to mirror the setting on the Uplink.

The spec allows the LNS leeway on setting the ADR bit, so this is not an entirely black & white issue fix.  If the Network temporarily cannot estimate the best data rate owing to rapid changes of the radio channel, it clears the ADR bit and thus signals the end-device that the end-device should rely on it's own behavior heuristics, i.e. usually set maximum power.  Our LNS code is probably not that smart yet.

JayKickliter: Good summary. Either that's new language in the spec or I missed it the last time I studied up on ADR.

JayKickliter (approved):

JayKickliter (approved):
Please squash before merging.

michaeldjeffrey (approved):

mikev:
The end-device on the Uplink sets the ADR bit. On the Downlink Ack we must (per the spec) also set the ADR bit to mirror the setting on the Uplink.

The spec allows the LNS leeway on setting the ADR bit, so this is not an entirely black & white issue fix. If the Network temporarily cannot estimate the best data rate owing to rapid changes of the radio channel, it clears the ADR bit and thus signals the end-device that the end-device should rely on it's own behavior heuristics, i.e. usually set maximum power. Our LNS code is probably not that smart yet.

PR #514, opened 2021-11-17
Macpie/img update

macpie:


dpezely (approved):
Will let Jay comment on the 2Q cache calls, but everything else looks good... with some optional suggested bits for Dockerfiles.

PR #518, opened 2021-11-20
Macpie/2021 11 20

macpie:


dpezely (approved):

macpie:
We need to get https://github.com/helium/blockchain-core/pull/1119 merged first

src/device/router_device_cache.erl
```diff
@@ -49,7 +52,7 @@ get(DeviceID) ->
 
 -spec get_by_devaddr(binary()) -> [router_device:device()].
 get_by_devaddr(DevAddr) ->
-    MS = ets:fun2ms(fun({_, D}) when D#device_v5.devaddr == DevAddr -> D end),
+    MS = ets:fun2ms(fun({_, D}) when D#device_v6.devaddr == DevAddr -> D end),
```
michaeldjeffrey, line 55: Maybe inline this
https://learnyousomeerlang.com/ets#the-concepts-of-ets (search for "Don't Drink Too Much Kool-Aid")

michaeldjeffrey: Interesting, last paragraph of Example 1
https://www.erlang.org/doc/man/ms_transform.html#example-1

The only `ms_transform` thing I can find is inside of `recon`. I don't see anywhere it's being imported. But I wonder if that's why we can use it here.

dpezely (approved):

PR #526, opened 2021-11-30
XOR filter test investigation

michaeldjeffrey:
To end with, I ran the whole suite 5 times without any failures.

Some tests (xor in particular) rely on being to treat Console endpoints
like a stable store. When we're paginating the results put all the
devices back so we can do it again.

### DIFF
--- a/test/console_callback.erl
+++ b/test/console_callback.erl
@@ -76,21 +76,23 @@ handle('GET', [<<"api">>, <<"router">>, <<"devices">>], Req, Args) ->
                         }
                     ],
                     undefined};
-            [{devices, {Qs, []}}] ->
-                true = ets:delete(Tab, devices),
+            [{devices, {Qs, [], Refill}}] ->
+                true = ets:insert(Tab, {devices, Refill}),
                 {[], undefined};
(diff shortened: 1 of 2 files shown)


test/console_callback.erl
```diff
@@ -76,29 +76,28 @@ handle('GET', [<<"api">>, <<"router">>, <<"devices">>], Req, Args) ->
                         }
                     ],
                     undefined};
-            [{devices, {Qs, []}}] ->
-                true = ets:delete(Tab, devices),
+            [{devices, {Qs, [], Refill}}] ->
+                true = ets:insert(Tab, {devices, Refill}),
```
dpezely, line 80: A `delete` becomes an `insert`...?

Would you add a bit more context to the PR's description regarding nature of these changes, please?

michaeldjeffrey: I put those sorts of explanations in commit messages.

> Some tests (xor in particular) rely on being to treat Console endpoints
like a stable store. When we're paginating the results put all the
devices back so we can do it again.

mikev (approved):

dpezely (approved):

macpie:
thanks for that one!

Issue #464, opened 2021-10-26
In Device Event Log "Export JSON" button downloaded JSON, add raw_packet for all transmissions up and down


buzzware:
1) Join requests have  "payload": "", "payload_size": 0 - should have the raw packet payload
2) Join-Accepts have a payload I can paste into https://lorawan-packet-decoder-0ta6puiniaut.runkit.sh - thats good
3) uplinks have the data payload, but not the whole packet, so I can't use https://lorawan-packet-decoder-0ta6puiniaut.runkit.sh

The packet decoder web tool is extremely useful, and for debugging I would like access to the entire raw data, which currently isn't in the JSON.

It would also be great if the console itself linked to a display of the output from lorawan-packet-decoder

macpie:
So you would like to see the full packet (join or uplink) in Console's debug tool? 

_Note that joins do not have a data payload, but we might just be naming thing differently._

buzzware:
If packets don't have a payload, I would use null, but null string is fine.
My main issue was when investigating #463, there were uplinks with a payload but not the whole raw packet which contains the header fields and is required to use the decoder tool.

PR #527, opened 2021-11-30
Macpie/issues/464

macpie:
https://github.com/helium/router/issues/464 This PR adds the `raw_payload` (abse64 encoded) meaning the entire lora packets to data sent to integrations but also to Console (for debug mode).


michaeldjeffrey:
Do we want to add a `raw_payload_size` so people don't get confused which field `payload_size` refers to?

macpie:
> Do we want to add a `raw_payload_size` so people don't get confused which field `payload_size` refers to?
 What about `raw_packet` instead I think it actually makes more sense

dpezely (approved):

michaeldjeffrey (approved):

Issue #495, opened 2021-11-04
Production Console function decoder output shows error: no_context, status:error


leroyle:
1. Are you using Helium hosted Console or running your own open source Console?
  Hosted

2. If you are on Helium hosted Console, which url are you currently on? 
     Production shows the error, Staging does not
3. If you are experiencing bugs in Helium hosted Console, please specify your Organization name and Device ID (ex. 3483050b-284d-47e2-8eda-e087fc03a753).
   a0b70b92-1f9a-45ac-98d5-e38711da95ea

Please describe your issue:

- Earlier in the day my device was not able to join the Production console. The device debug view never did show Join or Join Accept. 
- Switched to staging console, connected uplinked as expected
- Back to Production, same "no response" in debug view.
- Reported same to support.

Sometime later my device was able to connect to production console.
I noticed every other "Uplink Integration Request", Decoded Payload was showing
```
"decoded": {
      "error": "no_context",
      "status": "error"
    },
```
- The alternating uplinks were being decoded correctly.
- switched to Staging console, issue was not present, no decoding errors noticed.
- The Device within the console was created pre V2, with no changes since V2
- Region US915, Helium original hotspot,  Texas USA

A sample run follows below.

The current function, also shown below is a copy of my original made this afternoon, but cut down considerably, problem still exhibits in this newer function although perhaps not as consistently alternating between success/failure.

Current Function:
```
// sample 
// Raw:  a740f01528013d0000
// Decoded: 5;1;1;0;15;21:2:61

// base64 hex: d4BQeyACQVQ=
// decoded hex bytes: 7780507b20024154
// Not sure about this decoding as I did // capture the origina: 
// Might be 3:5:2:0:5:123:2:AT
function Decoder(bytes, port) { 
var payload = 
{
"Leroy":"66"
}
payload.by0 = bytes[0]; 
payload.by1 = bytes[1]; 
payload.by2 = bytes[2]; 
payload.by3 = bytes[3]; 
payload.by4 = bytes[4]; 
payload.by5 = bytes[5]; 
payload.by6 = bytes[6]; 
payload.by7 = bytes[7]; 
payload.by8 = 0xff;

return payload;
}
```

Sample run using the above function. This a complete partial run, I've broken up the markup so I could add the GOOD/BAD notices to make it easier to scan.

**GOOD DECODE**
```
Uplink: Integration request Decoded Payload
Event Informationa few seconds ago

{
  "id": "b0464e3e-7528-49dd-863b-805eb42d0d32",
[...]

Issue #486, opened 2021-11-02
update erlang v8


jdgemm:
Seeing persistent, recurring issues believe updating to erlang v8 will address.

dpezely:
The V8 javascript engine has been upgraded to the latest *stable* release, which is 9.5.172.25.

This involved a minor update to our fork of `erlang_v8`.

The new `v8` binary has been uploaded to the S3 bucket, and we are testing this now along with other router updates.

Vagabond:
Note that we need an MacOS build from someone with a fast Mac to put in s3 as well.

someone would checkout 0.5.7 tag of erlang_v8, `make -C c_src build` then check the tests all pass with `make test` and if they're good make a package using `make -C c_src release-tarball` and sending me the `erlang_v8-0.5.7-Darwin-x64.tar.gz` file

cc @michaeldjeffrey

dpezely:
Added instructions from the preceding comment to PR 4 of `erlang_v8` repo (our fork).

We should consider x86-64 vs amd64 versions of Macs, but that would be a new feature of `../erlang_v8/c_src/Makefile`.  It looks like there are scripts relying upon python2, which is beyond end-of-life officially, so deeper excavation must be done there.  Maybe address both at the same time?  (Instructions within that PR can serve as a jumping-off point for both aspects.)

dpezely:
Although we updated V8 once for this ticket, Google is on a 6 week cadence with releases.  Another `stable` release is available and tested locally to work fine: all our tests passed with the following change to our Makefile:

```makefile
V8_VERSION := "9.6.180.12"
```
However, we're close enough to the *next* release (that will presumably also be tagged `stable` by them), so let's update again then.

PR #528, opened 2021-11-30
Graceful recovery of worker after javascript errors

dpezely:
When a custom decoder function (written in JavaScript) fails, this PR has router track that it failed and avoids further calls to V8 VM for it.

An Erlang process still gets created, because overall, that leads to less computation at runtime when subsequent frames arrive to be passed through that custom decoder.

(There may be room for improvement.  It may be possible to terminate the Erlang process and test earlier in the work-flow when subsequent frames arrive yet before attempting to message the process.)

This partially addresses the reason why issue  #486 was created.

This fixes issue #495 for multiple causes that lead to why `no_context` was being returned within `decoded` object for custom decoders.

### DIFF
--- a/src/decoders/router_decoder_custom_sup.erl
+++ b/src/decoders/router_decoder_custom_sup.erl
@@ -91,18 +91,9 @@ add(Decoder) ->
     add(ID, Function).
 
 -spec delete(binary()) -> ok.
--ifdef(TEST).
-delete(ID) ->
-    case ets:info(?ETS, size) of
-        undefined -> ok;
-        _ -> ets:delete(?ETS, ID)
-    end,
-    ok.
--else.
 delete(ID) ->
     true = ets:delete(?ETS, ID),
     ok.
--endif.
 
 -spec decode(
     Decoder :: router_decoder:decoder(),
--- a/src/decoders/router_decoder_custom_sup.erl
+++ b/src/decoders/router_decoder_custom_sup.erl
@@ -91,9 +91,18 @@ add(Decoder) ->
     add(ID, Function).
 
 -spec delete(binary()) -> ok.
+-ifdef(TEST).
+delete(ID) ->
+    case ets:info(?ETS, size) of
+        undefined -> ok;
+        _ -> ets:delete(?ETS, ID)
+    end,
+    ok.
+-else.
 delete(ID) ->
     true = ets:delete(?ETS, ID),
     ok.
+-endif.
 
 -spec decode(
     Decoder :: router_decoder:decoder(),
@@ -130,7 +139,7 @@ init([]) ->
 
 -spec add(DecoderID :: binary(), Function :: binary()) -> {ok, pid()} | {error, any()}.
 add(ID, Function) ->
-    case is_valid_decoder_function(Function) of
+    case contains_decoder_function(Function) of
         false ->
             {error, no_decoder_fun_found};
         true ->
@@ -152,7 +161,7 @@ add(ID, Function) ->
     {ok, pid()} | {error, any()}.
 start_worker(ID, Hash, Function) ->
     {ok, VM} = router_v8:get(),
-    Args = #{id => ID, vm => VM, function => Function},
+    Args = #{id => ID, vm => VM, function => Function, hash => Hash},
     case supervisor:start_child(?MODULE, [Args]) of
         {error, _Err} = Err ->
             Err;
@@ -220,8 +229,8 @@ get_oldest_decoder() ->
             end
     end.
 
--spec is_valid_decoder_function(binary()) -> boolean().
-is_valid_decoder_function(Function) ->
+-spec contains_decoder_function(binary()) -> boolean().
+contains_decoder_function(Function) ->
     case re:run(Function, ?DECODER_FUNCTION_REGEX) of
         nomatch ->
             false;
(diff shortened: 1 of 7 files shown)


src/decoders/router_decoder_custom_sup.erl
```diff
@@ -91,9 +91,18 @@ add(Decoder) ->
     add(ID, Function).
 
 -spec delete(binary()) -> ok.
+-ifdef(TEST).
```
michaeldjeffrey, line 94: Do we still need this test wrapper for `delete/1`?
I ran all eunits, and all the changed CTs without failure

src/decoders/router_decoder_custom_worker.erl
```diff
@@ -74,17 +74,45 @@ init(Args) ->
...
+        case init_context(VM, Function) of
+            {ok, Context} ->
+                TimerRef = erlang:send_after(?TIMER, self(), timeout),
+                #state{
+                    id = ID,
+                    vm = VM,
+                    context = Context,
+                    function = Function,
+                    timer = TimerRef,
+                    backoff = backoff:type(backoff:init(?BACKOFF_MIN, ?BACKOFF_MAX), normal)
```
michaeldjeffrey, line 87: None of the backoff stuff is being used, can we drop it?

src/decoders/router_decoder_custom_worker.erl
```diff
@@ -107,13 +135,32 @@ init(Args) ->
...
+%%             %% Execution reaching here implies V8 vm restarted recently
 %%             {Delay, Backoff1} = backoff:fail(Backoff0),
+%%             %% Refresh their Context and JS definition:
 %%             _ = erlang:send_after(Delay, self(), ?INIT_CONTEXT),
 %%             {reply, {error, invalid_context}, State1#state{
 %%                 context = undefined,
 %%                 backoff = Backoff1
 %%             }};
-%%         {error, _} = Error ->
+%%         {error, Reason0} = Error ->
```
michaeldjeffrey, line 146: Does this big ol' comment get to go away? 🙏

src/decoders/router_decoder_custom_worker.erl
dpezely: If we drop all of the backoff functionality, yes it will keep the code simple, and then yes we can remove those chunks of commented-out code.

(The original motivation for backoff was to avoid a tight loop to retry creating a context when that initial call fails.  However, now with better understanding, a failed context should be sufficiently rare that there's little value gained from using the backoff library here.)

src/decoders/router_decoder_custom_worker.erl
```diff
@@ -228,10 +280,16 @@ decode(
...
+            {ok, Context1} = init_context(VM, Function),
             %% TODO: use Ferd's `backoff' library.
             %% See stateful use: ../device/router_device_channels_worker.erl
             decode(Payload, Port, UplinkDetails, State#state{context = Context1}, Retry - 1);
+        {error, Err} ->
+            %% TODO this eliminates any tolerance for intermittent
+            %% errors such as an occasional payload crashing their
+            %% JavaScript code, so maybe add `js_error_count` in
+            %% State and a threshold test here.
+            {{js_error, Err}, State#state{function = <<>>}};
```
michaeldjeffrey, line 292: This one's tricky. The code didn't crash the VM, and is still valid because it was able to be loaded.
But every call after this type of failure will give them an error `ignoring_invalid_javascript`. 
It might be useful to check if the `Err` is a binary, then take up the first `\n`.
That would get shoved into the `decoded.error` key of what we send to their integration.

test/router_decoder_custom_sup_SUITE.erl
```diff
@@ -0,0 +1,113 @@
...
+%%--------------------------------------------------------------------
+all() ->
+    [
+        v8_recovery_test
+    ].
+
+%%--------------------------------------------------------------------
+%% TEST CASE SETUP
+%%--------------------------------------------------------------------
+init_per_testcase(TestCase, Config) ->
```
michaeldjeffrey, line 34: I might not make a huge difference for 1 test, but we might be able to forego `init_per_testcase` and directly startup the `router_decoder_custom_sup` supervisor, since we don't touch anything else in the system.

src/decoders/router_decoder_custom_sup.erl
dpezely: Fixed in commit f814c72

src/decoders/router_decoder_custom_worker.erl
dpezely: Fixed in commit 731a85b

src/decoders/router_decoder_custom_worker.erl
dpezely: Fixed in commit 731a85b

test/router_decoder_custom_sup_SUITE.erl
dpezely: I omitted `init_per_testcase()` in an earlier iteration between commits. There was a lot of boilerplate code such as starting ETS, lager, etc., so reusing the existing pattern seemed more clean in the end.  (That earlier attempt is why there was the `ifdef(TEST)` wrapped around `delete()` that has since been removed.

Also, there will likely be another test added now for more nuanced error reporting back to the console, so the human doesn't just see the `ignoring_invalid_javascript` atom.

src/decoders/router_decoder_custom_worker.erl
dpezely: ok. Let me add a test exercising the reporting of errors.  (Adding the extraction up to the first newline didn't crash any existing tests.)

dpezely: Addressed in 772469b, but I leave it to you for resolving this comment thread-- whether that commit is sufficient or not.

test/router_decoder_custom_sup_SUITE.erl
dpezely: Managed to work the additional test case into an existing test; see commit 772469b

macpie (approved):

PR #538, opened 2021-12-06
Add EU868 payload size map

fishpepper:
It looks like the US table is used for EU868 as well. 
Reading
https://lora-developers.semtech.com/documentation/tech-papers-and-guides/the-book/packet-size-considerations
this should be different for EU868.


michaeldjeffrey (approved):
:pray: awesome thank you. And even put refs in the commit message

PR #539, opened 2021-12-06
Update issue templates

macpie:


michaeldjeffrey:
Can we add which router/console they're device is registered in?

dpezely (approved):
This seems sufficient for the template, as we don't what to overdo the upfront questions.

michaeldjeffrey (approved):

Issue #531, opened 2021-12-01
Add RxDelay to LoRaWAN Configs


lthiery:
Some users experience latency issues using Helium-hosted Router/Console and/or cell-based gateways. The default RX_DELAY of 1 second might be too tight. If a user could configure RX_DELAY, this would allow them to improve their change of receiving a downlink in time.

jdgemm:
hi @lthiery can you confirm what the max range allowable would be on the miner side?

michaeldjeffrey:
If there's a chance Miner's will drop packets on the ground if the delay is not what they're expecting, it would be nice for router to get notified somehow that something was wrong with the timestamp.

lthiery:
> If there's a chance Miner's will drop packets on the ground if the delay is not what they're expecting, it would be nice for router to get notified somehow that something was wrong with the timestamp.

I think it would be nice if the Miner/LightGateway responded with the [packet forwarder's error message (or tx_ack with no error, or timeout on waiting for tx_ack)](https://github.com/Lora-net/packet_forwarder/blob/master/PROTOCOL.TXT#L422). These all happen in case of an issue but are not responded back to OUI operator who requested the transmit:

```
 Value             | Definition
:-----------------:|---------------------------------------------------------------------
 NONE              | Packet has been programmed for downlink
 TOO_LATE          | Rejected because it was already too late to program this packet for downlink
 TOO_EARLY         | Rejected because downlink packet timestamp is too much in advance
 COLLISION_PACKET  | Rejected because there was already a packet programmed in requested timeframe
 COLLISION_BEACON  | Rejected because there was already a beacon planned in requested timeframe
 TX_FREQ           | Rejected because requested frequency is not supported by TX RF chain
 TX_POWER          | Rejected because requested power is not supported by gateway
 GPS_UNLOCKED      | Rejected because GPS is unlocked, so GPS timestamp cannot be used 
```

lthiery:
> hi @lthiery can you confirm what the max range allowable would be on the miner side?

I can't find a hard limit other than 255 seconds because that's how big the field is. I would suggest constraining it to 1-5 seconds so that we don't exceed the Join Delay.

lthiery:
Also, generally recall that RX2 hangs off of RX1. As per the regional spec, for all regions:
```
RECEIVE_DELAY2 = (SHALL be RECEIVE_DELAY1 + 1s)
```

buzzware:
[...]

PR #545, opened 2021-12-08
Increase RX1_DELAY and RX2_DELAY by 1s for internet latency tolerance

dpezely:
**DO NOT MERGE -- this should have been created as a DRAFT PR and is incomplete / not yet correct.**

This should resolve issue #531 since internet latency seems to be a contributing factor there.

This increases the defaults for all regions because router lacks any configuration item specifying the region within which it's deployed.

This PR does NOT make this value configurable yet, as that feature hasn't been merged for Console.

Added comments within the source code to assist when needing to correlate the LoRaWAN spec with the code and vice versa.

### DIFF
--- a/src/device/router_device_routing.erl
+++ b/src/device/router_device_routing.erl
@@ -88,7 +88,13 @@
 
 %% Replay
 -define(REPLAY_ETS, router_device_routing_replay_ets).
--define(RX2_WINDOW, timer:seconds(2)).
+%% From LoRaWAN Link Layer 1.0.4 Spec, §5.7 Setting Delay between TX and RX:
+%% "RX2 always opens 1s after RX1," and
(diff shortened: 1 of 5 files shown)


lthiery:
I believe you need to also update the RX_DELAY field in the JoinAccept frame.

It is currently set to 0, in accordance with [this table](https://github.com/helium/router/blob/8a103868831e99f72600ba34ef2f8ea2c3351e51/include/lorawan_vars.hrl#L10). Note: the specification of the `RXTimingSetupReq` byte matches the specification of the RX_DELAY byte in JoinAccept.

image

dpezely:
@lthiery Thanks for calling that out.  I'm glad your eyes & brain are on this!

I should have marked this PR as `DRAFT` since I'm still new enough and handling sections of code that are new to me.

dpezely:
Superseded by PR #548

Issue #458, opened 2021-10-19
Previous Azure integration clinging on


jdgemm:
A misc integration error appears for an Azure integration in a device event log even though that integration is no longer being used for that device.

Reproducible:

After removing azure integration from a device, adding another integration (request bin), then removing the new integration, after a number of uplinks the misc integration error appears

<img width="1191" alt="Screen Shot 2021-10-19 at 3 06 13 PM" src="https://user-images.githubusercontent.com/36315460/138002032-480186c9-61e8-4388-bee2-4a83b6aef5ba.png">
.

PR #547, opened 2021-12-10
Remove a channels backoff when it's deleted

michaeldjeffrey:
When removing channels because the API has told us they should no longer exist, make sure we remove the backoffs with them.

ref: #458


dpezely (dismissed):
Based upon your PR description, that's a fine bit of sleuthing you did there!

dpezely (approved):

Issue #535, opened 2021-12-04
CFList Type 1 has wrong endianness 


mkuyper:
It looks like the 16-bit channel mask values in the CFList my devices are receiving in US915 are not in little-endian format as required by the LoRaWAN specification:
```
PDU (enc): 20 3B42 5A36CF59 1606D739 A4 8E 75F7A66EFB3F2144F46FE5BBE31375B8 AF176C59
PDU (dec): 20 D35E C4486532 4C000048 08 00 FF000000000000000000000000000001 546D3D38

Join accept:
{'AppNonce': 12869331,
 'CFList': {'chmask': 'FF0000000000000000000000000000', 'type': 1},
 'DLSettings': 8,
 'DevAddr': 1207959628,
 'MHdr': 32,
 'MIC': 943549780,
 'NetID': 3302728,
 'RxDelay': 0,
 'statusMIC': True}
```
ChMask0 is thus `FF00` i.e. 0x00FF, enabling channels 0-7 and leaving the Helium channels of 8-15 disabled.

pvolodin:
And there seems to be no channel 65 ( 904.6 MHz / SF8 / BW500 )

mkuyper:
> And there seems to be no channel 65 ( 904.6 MHz / SF8 / BW500 )

Yes, that is correct. Not sure if that's intentional, but at least it's the same as is done via the LinkAdrReq command:
```
FOpts  : x0310000070031000FF00
         LinkADRReq<x03 u1(TXPow:0-3=0,DR:4-7=1) u2(ChMask:0-15=x0) u1(NbTrans:0-3=0,ChMaskCntl:4-6=7,RFU:7=0)>
         LinkADRReq<x03 u1(TXPow:0-3=0,DR:4-7=1) u2(ChMask:0-15=xFF00) u1(NbTrans:0-3=0,ChMaskCntl:4-6=0,RFU:7=0)>
```
The first command disables all channels (0-71), and the second enables channels 8-15.

pvolodin:
@mkuyper 
LinkADRReq controls datarate, CFList doesn't.
The sample you provided sets DR1, so there's no need in channel 65.

pvolodin:
@mkuyper  Can you please point me to the paragraph in LoRaWAN spec or regional settings doc which prohibits disabling the channels you don't need right now and in the foreseeable future? In US in most cases you cannot disable all 125 KHz channels as they shall be used for data rate back-off, but DR4 never can be DRNext.

[...]

PR #553, opened 2021-12-18
fix ct test for #535

mikev:
CT test is broken

%%% router_lorawan_SUITE ==> US915.lw_join_test: FAILED
%%% router_lorawan_SUITE ==>
Failure/Error: ?assertEqual([8,9,10,11,12,13,14,15], Mask)
  expected: [8,9,10,11,12,13,14,15]
       got: [0,1,2,3,4,5,6,7]
      line: 572

We changed the mask encoding from big-endian to little-endian.  This would change the mask from 0xF0 to 0x0F.  Thus the CT test should be modified to match 0..7.

### DIFF
--- a/test/router_lorawan_SUITE.erl
+++ b/test/router_lorawan_SUITE.erl
@@ -568,7 +568,7 @@ lw_join_test(Config) ->
         {channel_mask, Mask} ->
             case Region of
                 'US915' ->
-                    ExpectedMask = lists:seq(8, 15),
+                    ExpectedMask = lists:seq(0, 7),
                     Mask = ExpectedMask;
                 _ ->
                     ct:pal("Mask is ~p", [Mask]),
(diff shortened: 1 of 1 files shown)


test/router_lorawan_SUITE.erl
```diff
@@ -568,7 +568,7 @@ lw_join_test(Config) ->
         {channel_mask, Mask} ->
             case Region of
                 'US915' ->
-                    ExpectedMask = lists:seq(8, 15),
+                    ExpectedMask = lists:seq(0, 7),
                     Mask = ExpectedMask;
```
dpezely, line 572: Not that you touched this line, but for sanity when debugging this test case in future it would be useful to see both values in test output instead of `bad match` or whatever.

```suggestion
                    ?assertEqual(ExpectedMask, Mask);
```

dpezely (dismissed):

macpie (approved):

PR #557, opened 2021-12-21
RxDelay Part 2

dpezely:
This expands upon PR #531

- Expand rx_window() to use RxDelay

PR #548, opened 2021-12-14
Add `rx_delay` from Console to LoRaWAN config

dpezely:
- This should resolve issue #531
- This will assist devices in regions where internet backbone latency contributes to timing issues; e.g., AU->US
- For this feature to be fully function, it is blocked by Console [PR#952](https://github.com/helium/console/pull/952)

PR #563, opened 2021-12-29
Accommodate change to `rx_delay` from Console without device re-joining

dpezely:
- This is Part 3 of adding `rx_delay` feature, the value of which may be set in [Console](https://github.com/helium/console)
- Part 2 was [PR#557](https://github.com/helium/router/pull/557)
- Part 1 was [PR#548](https://github.com/helium/router/pull/548)
- This set of PRs should fully resolve issue #531

### DIFF
--- a/src/device/router_device_worker.erl
+++ b/src/device/router_device_worker.erl
@@ -359,12 +359,21 @@ handle_cast(
         {ok, APIDevice} ->
             lager:info("device updated: ~p", [APIDevice]),
             router_device_channels_worker:refresh_channels(ChannelsWorker),
+            RxDelayOld = maps:get(rx_delay, router_device:metadata(Device0), 0),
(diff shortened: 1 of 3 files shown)


macpie (approved):

src/device/router_device_worker.erl
```diff
@@ -1651,6 +1660,8 @@ handle_frame_timeout(
                 },
                 Device0
             ),
+            %% Handle sequential changes to rx_delay: prev was ack'd, now a potentially new RxDelay
+            RxDelay = maps:get(rx_delay, router_device:metadata(Device0), 0),
```
michaeldjeffrey, line 1664: Pulling `RxDelay` all the way out of the `case` causes the next line to expect the binding to match.

dpezely: Good catch.  I must have messed that up when resolving the earlier conflict from merging `master` into this branch.

macpie (approved):

PR #567, opened 2022-01-04
Add grpc connections count metric

michaeldjeffrey:
We could do the same thing with supervisor:which_children, but the
acceptor_pool uses a different way to report workers than investigating
the supervision tree.


dpezely (approved):

macpie:
Can we add a test in `router_metrics_SUITE` ?

macpie (approved):

Issue #562, opened 2021-12-24
Data shows passing through hotspot in console but not in explorer


Destable:
**Describe the bug**
Sensor is successfully sending packets through hotspot Mythical Maroon Bat according to console. But in the data tab for that hotspot in explorer shows no recent data. This has gone on since sensor activation 5 days ago. 

![7CFD4098-B281-448B-A464-1574EC01C47F](https://user-images.githubusercontent.com/96600723/147300539-4d9172cc-fb9f-40e3-959e-21ac7e7d8130.jpeg)
![4A853610-43F0-4D6B-8113-A9E206576511](https://user-images.githubusercontent.com/96600723/147300543-973ee404-4a6e-4bf2-b81b-8259fafcf14e.jpeg)

**To Reproduce*

images of console showing the packet transfers attached. Image of the same hotspots explorer not showing the data attached.

**Expected behavior**
I expected that after some reasonable delay, all data packets the transferred through mythical maroon bat would also show in explorer.

**Device Info (please complete the following information):**
 - Environment (Production, Staging) : production I think, I’m still fairly new. I’m not sure what the difference is but I signed up for Console and started using it.
 - Console Device ID: 
2baa9cd8-d6a3-4443-8b0d-66f2bce31c06
 - Device dev eui: A840418CF182C0BF
 - App eui: A000000000000107
 - How often is device sending data: Once per day unless triggered
 - Region: USA
 - Hotspots (names) used to transfer data: Mythical Maroon Bat

*NOTE: do not post your device `app_key`*

**Additional context**
According to console it looks like my hotspot is the only one receiving data from my sensor. At first I assumed there was a lag and that’s why I was not seeing it appear under explorer for my hotspot, but after so many days I reported it on Discord. Cyborg on discord asked me to post it here..

**Screenshots**
Screenshots are attached, but may be above in the initial problem description.

ahmadsayed:
Getting similar issue, the data I sent does not reported on the explorer of any hotspots showed on in the uplink or acknowledge.

It only shows the data I sent 7 days ago, any data I sent during this week and today , does not appear in the explorer, while it appears int he console, and it successfully received by mqtt server and my app.

hotspots - not mine - 
Tangy Saffron Orangutan
Sour Blush Ferret

jdgemm:
Thanks for reporting. Currently, the team is out on winter holidays for much needed rest, but when they return this will be a high priority item to investigate.

anders462:
[...]

PR #568, opened 2022-01-05
Macpie/close submit

macpie:


macpie:
Fix for https://github.com/helium/router/issues/562

include/metrics.hrl
```diff
@@ -47,6 +48,8 @@
     {?METRICS_VM_CPU, prometheus_gauge, [cpu], "Router CPU usage"},
     {?METRICS_VM_PROC_Q, prometheus_gauge, [name], "Router process queue"},
     {?METRICS_VM_ETS_MEMORY, prometheus_gauge, [name], "Router ets memory"},
-    {?METRICS_XOR_FILTER, prometheus_gauge, [], "Router XOR Filter udpates"},
+    {?METRICS_XOR_FILTER, prometheus_counter, [], "Router XOR Filter udpates"},
```
michaeldjeffrey, line 51: This is gauge because the metric is the cost in DC of an XOR txn.

macpie: Easier to read I think if a counter where it only increment with a new submit, no?

src/metrics/router_metrics.erl
```diff
@@ -127,7 +128,12 @@ ws_state(State) ->
 
 -spec xor_filter_update(DC :: non_neg_integer()) -> ok.
 xor_filter_update(DC) ->
-    _ = prometheus_gauge:set(?METRICS_XOR_FILTER, DC),
+    _ = prometheus_counter:inc(?METRICS_XOR_FILTER, DC),
```
michaeldjeffrey, line 131:
```suggestion
    _ = prometheus_counter:inc(?METRICS_XOR_FILTER, [], DC),
```

I increment by the amount of DC spent.

michaeldjeffrey: This is already covered when the second argument is not a list.

michaeldjeffrey (approved):

PR #576, opened 2022-01-10
Replace MB entry with same shape

michaeldjeffrey:
otherwise ets will drop our update on the ground

### DIFF
--- a/src/device/router_device_routing.erl
+++ b/src/device/router_device_routing.erl
@@ -667,7 +667,7 @@ maybe_multi_buy(Offer, Attempts, Device) ->
     end.
 
 -spec lookup_mb(binary()) ->
-    {ok, binary(), non_neg_integer(), non_neg_integer()} | {error, not_found}.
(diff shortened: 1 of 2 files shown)


mikev (dismissed):

macpie (dismissed):

src/device/router_device_routing.erl
```diff
@@ -667,7 +667,7 @@ maybe_multi_buy(Offer, Attempts, Device) ->
     end.
 
 -spec lookup_mb(binary()) ->
-    {ok, binary(), non_neg_integer(), non_neg_integer()} | {error, not_found}.
+    {ok, binary(), non_neg_integer(), integer()} | {error, not_found}.
 lookup_mb(PHash) ->
```
dpezely, line 671: The text of your commit message might be useful as a code comment, but perhaps there's a better place to put it than here?

```suggestion
%% use -1 to mean deny more
lookup_mb(PHash) ->
```

dpezely (dismissed):

dpezely (approved):

PR #578, opened 2022-01-11
Upgrade hackney to 1.18.0

macpie:


dpezely (dismissed):

src/channels/router_channel_utils.erl
```diff
@@ -155,49 +155,49 @@ url_test_() ->
     [
         ?_assertEqual(
             {
-                <<"https://website.com/">>,
+                <<"https://website.com">>,
```
dpezely, line 158: FYI: while it's not new code and just a change, it may still be worth noting:

The canonical placeholder domain names are `example.com` and `.org`, `,net` per RFC 2606.  If any such host names are inadvertently used in a DNS lookup, there's no harm, yet a dummy HTTP server will answer (probably with 404).

https://datatracker.ietf.org/doc/html/rfc2606#section-3

dpezely (approved):

michaeldjeffrey (approved):

PR #585, opened 2022-01-26
OTP 24 Support

macpie:


macpie:
I encourage everyone to checkout this branch and run tests locally, with docker if you don't feel like installing OTP24 on your machine just yet

dpezely:
Merging the `macpie/otp-24` branch into `dpezely/fix-rx_delay-for-join-accept` and updating to OTP-24.2 locally on Linux (xubuntu 24.04.3)
- `lw_join_test` fails for all regions (US, EU, CN, etc.)
- all other tests passed
- Same outcome when running `make docker-build docker-test`

src/channels/router_channel_utils.erl
```diff
@@ -155,49 +155,49 @@ url_test_() ->
...
-                <<"https://website.com/">>,
+                <<"https://website.com">>,
                 []
             },
             make_url("https://website.com", [], #{})
         ),
         ?_assertEqual(
             {
-                <<"https://website.com/?one=two">>,
+                <<"https://website.com?one=two">>,
```
dpezely, line 165: Removing the slash makes the string no longer be a valid URL.

If this is truly intentional, please add a comment noting the fact.

This applies to multiple instances, below this line.

src/device/router_device_routing.erl
```diff
@@ -666,8 +666,7 @@ maybe_multi_buy(Offer, Attempts, Device) ->
             end
     end.
 
--spec lookup_mb(binary()) ->
-    {ok, binary(), non_neg_integer(), integer()} | {error, not_found}.
+-spec lookup_mb(binary()) -> {ok, binary(), non_neg_integer(), integer()} | {error, not_found}.
```
dpezely, line 669: FYI: when there are multiple items within a return value like this, Dialyzer accommodates adding variable names to the result type-- just like with function arguments.

Let's try doing this for new & modified code while it's fresh in our minds.

src/device/router_device_worker.erl
```diff
@@ -2075,10 +2081,39 @@ save_and_update(DB, CF, Pid, Device) ->
...
+            lager:info("got device: ~p", [APIDevice]),
+            RxDelayOld = maps:get(rx_delay, router_device:metadata(Device), 0),
+            RxDelayNew = maps:get(rx_delay, router_device:metadata(APIDevice), RxDelayOld),
+            Metadata =
+                case RxDelayOld == RxDelayNew of
+                    true ->
+                        router_device:metadata(APIDevice);
+                    false ->
+                        maps:put(new_rx_delay, RxDelayNew, router_device:metadata(APIDevice))
+                end,
```
dpezely, line 2107: Depending upon which gets merged first:

This expression for populating `Metadata` and where it's used below will create a merge conflict if `dpezely/fix-rx_delay-for-join-accept` gets merged to master first.

In that case, remove this entire `Metadata` statement and variable, and where it gets used: that expression gets replaced by a single `maps:merge()` call.

Makefile
```diff
@@ -19,6 +19,16 @@ test: | $(grpc_services_directory)
...
 	$(REBAR) dialyzer
 
+# This will run all ct exepct the lora suite that for now cannot be ran on MacOS
+test-mac: | $(grpc_services_directory)
+	$(REBAR) fmt --verbose --check rebar.config
+	$(REBAR) fmt --verbose --check "{src,include,test}/**/*.{hrl,erl,app.src}" --exclude-files "src/grpc/autogen/**/*"
+	$(REBAR) fmt --verbose --check "config/{test,sys}.{config,config.src}"
+	$(REBAR) xref
+	$(REBAR) eunit
+	$(REBAR) ct --suite=router_SUITE,router_channel_aws_SUITE,router_channel_azure_SUITE,router_channel_console_SUITE,router_channel_http_SUITE,router_channel_mqtt_SUITE,router_channel_no_channel_SUITE,router_console_api_SUITE,router_console_dc_tracker_SUITE,router_data_SUITE,router_decoder_SUITE,router_decoder_custom_sup_SUITE,router_decoder_custom_worker_SUITE,router_device_channels_worker_SUITE,router_device_devaddr_SUITE,router_device_routing_SUITE,router_device_worker_SUITE,router_discovery_SUITE,router_downlink_SUITE,router_grpc_SUITE,router_metrics_SUITE,router_sc_worker_SUITE,router_v8_SUITE,router_xor_filter_SUITE
```
dpezely, line 29: That will be too easily overlooked and not updated in future.

How about omitting just the non-`_SUITE.erl` files and those without `_lorawan_` in the name with a bit of old school cartoon swearing:

```suggestion
	$(REBAR) ct --suite=$(cd test/ && ls -1N *_SUITE.erl | grep -v _lorawan_ | awk '{gsub(".erl$", ""); printf("%s,", $0)}')
```

dpezely (dismissed):
OTP-24 !!!

src/channels/router_channel_utils.erl
michaeldjeffrey: I guess not anymore?
https://stackoverflow.com/a/42193734/1921234

dpezely: So many RFCs, so little time...

Makefile
```diff
@@ -19,6 +19,16 @@ test: | $(grpc_services_directory)
...
 	$(REBAR) dialyzer
 
+# This will run all ct exepct the lora suite that for now cannot be ran on MacOS
+test-mac: | $(grpc_services_directory)
+	$(REBAR) fmt --verbose --check rebar.config
+	$(REBAR) fmt --verbose --check "{src,include,test}/**/*.{hrl,erl,app.src}" --exclude-files "src/grpc/autogen/**/*"
+	$(REBAR) fmt --verbose --check "config/{test,sys}.{config,config.src}"
+	$(REBAR) xref
+	$(REBAR) eunit
+	$(REBAR) ct --suite=$(cd test/ && ls -1N *_SUITE.erl | grep -v _lorawan_ | awk '{gsub(".erl$", ""); printf("%s,", $0)}')
```
michaeldjeffrey, line 29: This command doesn't run with the `ls` that comes with macOS. `-N` is not a valid option

dpezely: Grr.... I keep forgetting that macos uses an older base from FreeBSD, which doesn't support all of these flags that different Linux distros keep adding.

(The `-N` flag is to negate quoting each filename, which shouldn't be used here anyway since we don't have spaces in these filenames.)

Also, I realized that I missed a slash in the reg-ex.

```suggestion
 	$(REBAR) ct --suite=$(cd test/ && ls -1 *_SUITE.erl | grep -v _lorawan_ | awk '{gsub("\.erl$", ""); printf("%s,", $0)}')
```

PR #595, opened 2022-02-01
Improve CI/CD

macpie:


michaeldjeffrey (approved):

.github/workflows/main.yml
```diff
@@ -0,0 +1,144 @@
...
+        uses: actions/download-artifact@v2
+        with:
+          name: build
+      - name: untar
+        run: |
+          mkdir _build
+          tar -xvf build.tar.gz -C _build
+      - name: dialyzer
+        run: ./rebar3 dialyzer
+  suites:
```
macpie, line 100: @michaeldjeffrey can we put this into the build process instead of being another step as it only takes few seconds?

michaeldjeffrey: Tried and failed. Will try again when we have even more GHA experience

PR #600, opened 2022-02-03
Join-Accept implies `rx_timing_setup_ans` for `RXDelay`

### LINKED ISSUE #597: RxDelay set in the console is not propagated in the JoinAccept
**Describe the bug**
The value of the RxDelay set in the console is not sent to the device in the JoinAccept.

**To Reproduce**
1. Create a device profile with an RxDelay > 1. In my case CFList and ADR are set
2. Create a new device with this profile
3. Start the device an monitor the join accept
4. The join accept will be like: `E929A1 486532 0E080048 08 00 00FF0000000000000000000000000001` where RxDelay is set to the the [router default value of 0](https://github.com/helium/router/blob/8a103868831e99f72600ba34ef2f8ea2c3351e51/include/lorawan_vars.hrl#L10) (13th byte before the CFList).

**Expected behavior**
The value set in the console is sent in the joinAccept

**Device Info (please complete the following information):**
 - Environment (Production, Staging): both
 - Console Device ID: 953f8b66-9f13-4df4-bbc4-881ccc4f8aa0 (Prod) - 6dc0f3fa-cb37-4fa9-bb17-e19b80676fe1 (US)
 - How often is device sending data : Every 15 min for the device in prod. Once in a while for the device on staging
 - Region: EU868 (Prod) - US915 (Staging)
 - Hotspots (names) used to transfer data:  real-smoke-nightingale (Prod) - great-plum-mallard (Staging)

**Additional context**
The RxDelay was correctly set when I opened issue #571 and I belive it was working Moday on staging when I commented on #582 but I did not kept the logs to check again

dpezely:
LoRaWAN Join-Accept implies `rx_timing_setup_ans` because `RXDelay` was in the Join-Accept.

- 6.2.6 Join-Accept frame, which includes RXDelay
- This should resolve #597
- Starting this PR as a Draft because the test framework may also need to be corrected

LoRaWAN Join-Accept implies `rx_timing_setup_ans` because `RXDelay` was in the Join-Accept frame, but router's Join logic doesn't explicitly set that answer.

That is, the ACK from the device is equivalent to `rx_timing_setup_ans` in FOpts in a frame (after join-accept), but FOpts are omitted on Join.


dpezely:
We confirmed `RXDelay` value in Join-Accept header via [virtual-lorawan-device](https://github.com/helium/virtual-lorawan-device) and [lorawan-sniffer](https://github.com/helium/lorawan-sniffer) as well as handling `rx_timing_setup_req` & `rx_timing_setup_ans` in FOpts after join.

PR #624, opened 2022-02-16
If port is missing in MQTT integration lets default to 1883

macpie:


dpezely (approved):

src/channels/router_mqtt_channel.erl
```diff
@@ -429,20 +429,20 @@ connect(URI, DeviceID, Name) ->
     case uri_string:parse(URI) of
         #{
             host := Host,
-            port := Port,
             scheme := Scheme
         } = Map when
             Scheme == <<"mqtt">> orelse
                 Scheme == <<"mqtts">>
         ->
+            Port = maps:get(port, Map, 1883),
```
michaeldjeffrey, line 437: Should we default to `8883` if the scheme is `mqtts`?

macpie: Good idea!

michaeldjeffrey (approved):

dpezely (approved):

PR #630, opened 2022-02-18
correct join channels for AS923 variants

mikev:
The Region Plan AS923 variants "1", "2", "3" and "4" define different and unique fixed Join channels.  The correct Join channels can be viewed in the Miners sys.config:
https://github.com/helium/miner/blob/master/config/sys.config#L119

Fix the Router's values to the correct values.


macpie:
is This rdy for review?

macpie (approved):

Issue #611, opened 2022-02-10
Advances JSON Message Template not including DC Balance


takeabyte:
When trying to pass the DC Balance to an integration Endpoint it will show the correct amount of DC in the Output Preview Window  next to the Template Body, however it will not be sent out to the Endpoint, but instead show the Inverted section, which means the key doesn't exist, is false, or is an empty list.
 

```
{{#dc}}
"dc_balance": "{{balance}}",
"nonce": "{{nonce}}",
{{/dc}}
{{^dc}}
  "No DC Found",
{{/dc}}
```

takeabyte:
unfortunately with current console v.2.2.5 DC Balance still won't show up on the Endpoint. 
Expected output should be
`"category": "uplink",
    "data": {
      "dc": {
        "balance": 914118,
        "nonce": 2,`

Advanced JSON Template in Use: 
`{
"reported_at": {{reported_at}},
{{#dc}}
"dc_balance": "{{balance}}",
"nonce": "{{nonce}}"
{{/dc}}
}`

Result coming back from integration:
`{
"reported_at": 1648160197185,
"dc_balance": ,
"nonce": 2
}`

michaeldjeffrey:
> Advanced JSON Template in Use: `{ "reported_at": {{reported_at}}, {{#dc}} "dc_balance": "{{balance}}", "nonce": "{{nonce}}" {{/dc}} }`

Can you try with the key `{{ dc_balance }}`?
https://github.com/helium/router/blob/master/src/device/router_device_channels_worker.erl#L667-L670

macpie:
The current key is `dc_balance`, mistake on our part we will update it to `balance` asap.

takeabyte:
yes. that is working for now. good you will change it soon, also to be in line with the Advanced JSON testoutput.

macpie:
@takeabyte  This should be corrected in `2022.03.16.2` and was just pushed

takeabyte:
yes. all good now. Thanks!

PR #655, opened 2022-03-17
Send DC balance to channels (#611)

barndon:
This PR adds DC balance and nonce (integers, both) to channel messages.  The added fields look like this:

```
   "dc" : {
      "dc_balance": 42,
      "nonce": 123456
   }
```


macpie:
Fix for #655

dpezely (approved):

Issue #633, opened 2022-02-24
LinkAdrReq can be set to DR4 on Fixed Channel Plans when CFList Disabled


lthiery:
In Fixed Channel Plans (such as US915 and AU915), when CFList is disabled, ChMask selecting the Helium Network's subband is communicated via LinkAdrReq. We use the multiple LinkAdrReq approach as described here
![image](https://user-images.githubusercontent.com/1606640/155587439-26156290-69e2-47b7-95be-633ad2418243.png)

When a JoinRequest and the subsequent Uplink uses 500 MHz, it can compel Router to provide a LinkAdrReq with DR set to 4:
![image](https://user-images.githubusercontent.com/1606640/155587510-aeacf561-d29e-4f39-ae99-ef98e19442ba.png)

Since Channel 65 is disabled in the ChMask, setting DR to 4 is incompatible; DR4 is not valid on CH8-15.

Note, a JoinRequest and subsequent Uplink using 125 MHz does not compel DR to be set to 4:
![image](https://user-images.githubusercontent.com/1606640/155587901-4bc61c8c-745e-4ae7-abaa-6e604f1dd4a3.png)

Therefore, it appears that DR configuration in the LinkAdrReq is configured in response to the DR of the Uplink. This logic may be appropriate but must not exceed DR3. Another approach would be to consider enabling CH65 which would make DR4 possible.

mikev:
@lthiery - Can you link to a full json and csv packet trace?  Ideally the trace will include the join request and include a portion where FCnt starts at zero.

lthiery:
[dr4-test.csv](https://github.com/helium/router/files/8136380/dr4-test.csv)

You can replay the above with the latest release of [LoRaWAN Sniffer utility](https://github.com/helium/lorawan-sniffer/releases/tag/v0.2.4).

It looks like this:
```
JoinRequest	914.2 MHz 	DataRate(SF8, BW500)	RSSI: -112	LSNR: 5.5
	  AppEui: 35BEED137ACD344B DevEui: 653D0F09DDED4DB0 DevNonce: 9AD2
JoinAccept	927.5 MHz 	DataRate(SF7, BW500)
	AppNonce: 15664F NetId: 326548 DevAddr: 48000287
	DL Settings: DLSettings(8) RxDelay: 1
	Newskey: AES128([FF, 52, C8, 73, 26, 99, 9A, A6, DE, DF, 76, 59, C8, 4, 35, 5A])
[...]

PR #656, opened 2022-03-17
Ignore TxDr Logic

mikev:
Solution to address #633.  The spec allows Tx=0xF and Dr=0xF to indicated that DR and/or TX should be ignored and the end-device should continue to use it's current DR and TX settings.

### DIFF
--- a/src/device/router_device_worker.erl
+++ b/src/device/router_device_worker.erl
@@ -1153,7 +1153,7 @@ maybe_send_queue_update(Device, #state{queue_updates = {ForwardPid, LabelID, _}}
     ForwardPid !
         {?MODULE, queue_update, LabelID, router_device:id(Device), [
             downlink_to_map(D)
-            || D <- router_device:queue(Device)
+         || D <- router_device:queue(Device)
         ]},
     ok.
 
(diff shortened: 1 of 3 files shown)


src/device/router_device_worker.erl
```diff
@@ -1869,12 +1885,28 @@ channel_correction_and_fopts(Packet, Region, Device, Frame, Count, ADRAdjustment
                 %% rate in the form of "SFdd?BWddd?" so that's what
                 %% we'll give it.
                 NewDr = lorawan_mac_region:dr_to_datar(Region, NewDataRateIdx),
+                NewTxPowerIdx = 0,
```
macpie, line 1888: You are setting `NewTxPowerIdx` to 0 here but it is also coming from above line 1879. Note that dialyzer is complaining because ` 0 =:= NewTxPowerIdx::pos_integer()`

mikev: Fixed.

michaeldjeffrey (approved):

pvolodin:
This will not work for LoRaWAN 1.0.0 - 1.0.3 .

mikev:
1.0.4 changes are not communicated via a version and are therefore intended to be backwards compatible with prior versions.  That is my understanding after discussions with LoRaWAN spec authors.  In other words it is considered safe to implement 1.0.4 specs and they should be compatible with previous version.  An example of this is the DevNonce which as of 1.0.4 must be an incrementing monotonic value instead of a random value.

A 1.0.0 to 1.0.3 end-device will view the value 0xF as an unknown DR or TX index setting.  Therefore, the logical effect should be that the value is ignored.

PR #658, opened 2022-03-18
Fix formatting

macpie:


dpezely (approved):
GitHub diffs with Settings -> `hide whitespace` for the win!

macpie:
> GitHub diffs with Settings -> `hide whitespace` for the win!

I which it saved it so that I don't have to click every time...

PR #663, opened 2022-03-22
Handle case where aws thing has multiple certificates attached

macpie:
Note: the last certificate seems to be the oldest


michaeldjeffrey (approved):
Does that `lager:warning` print out with the device or channel id?

macpie:
> Does that `lager:warning` print out with the device or channel id?

They all do

dpezely (approved):

PR #667, opened 2022-03-24
Add ops runbook

dpezely:
"For regular operations and maintenance of the Router back-end of the Console"

There's a delicate balance of criteria when crafting a runbook like this:
- Accommodate new staff
- Accommodate community members running their own servers
- Accommodate someone having been woken in middle of sleep by a system alarm
- Be brief
- Be concise
- Be accurate
- Avoid leaking account information or operational security
- Avoid a level of detail that would require perpetual edits, because too granular means it *will* get out of date quickly

Good rules of thumb:
- New staff work from the runbook and revise as they discover something out of date.
- When on-call, work by the book *because the procedure may have changed* since prior on-call

### DIFF
--- a/docs/runbook.md
+++ b/docs/runbook.md
@@ -1,0 +1,245 @@
+Runbook
+=======
+
+For regular operations and maintenance of the Router back-end of the Console
+
+## Once Per New Server
+
+General installation instructions are available on the Helium documentation
+website.
+
+> Start here:
+[Run a Network Server](https://docs.helium.com/use-the-network/run-a-network-server/)
+which goes from `git clone` through to various initial `docker` commands
+such as loading an initial snapshot and testing with a real or virtual device.
+
+Syntax of **Bash** shell is used below.
+
+Create shell aliases.
+
+Router:
+
+```bash
+alias router='docker exec -it helium_router router'
+alias wallet='~/helium-wallet-rs/target/release/helium-wallet'
+```
+
+Console:
+
+```bash
+alias console_down='docker-compose down'
+alias console_psql='docker exec -it helium_postgres psql -U postgres'
+alias console_restart='console_down && console_up'
+alias console_up='docker-compose build && docker-compose up -d && docker-compose logs -f'
+alias console_update='docker-compose build --pull && console_restart'
+```
+
+Miscellaneous:
+
+```bash
+alias alert='notify-send --urgency=low -i "$([ $? = 0 ] && echo terminal || echo error)" "$(history|tail -n1|sed -e '\''s/^\s*[0-9]\+\s*//;s/[;&|]\s*alert$//'\'')"'
+```
+
+With the above set of aliases defined and available in your current shell,
+confirm:
+
+```bash
+router --help
+```
+
+## Stop/Start
+
+On production servers, Console versus Router run on different hosts to
+isolate and distribute workloads.
+
+Because staging has far less traffic, everything runs on one host.
+
+```bash
+docker-compose down
+docker-compose up -d
+```
+
+## Install A Release
+
+Published Docker images are
+[tagged](https://quay.io/repository/team-helium/router?tab=tags)
+as `production` or `staging` as appropriate.
+
+Install:
+
+```bash
+router install production
+```
+
+Restart the service:
+
+```bash
+docker-compose down && docker-compose up -d
+```
+
+## File System Cleanup
+
+Keep `blockchain` because it contains your swam key.
+
+Keep `router.db` because it contains device cache.
+
+Everything else under `/var/data/` may be purged for reclaiming space on the
+file system.
+
+## Funding
+
+The wallet address may be extracted from a running server instance:
+
+```bash
+router peer addr
+```
+
+Take only the long alphanumeric sequence from the results *without* `/p2p/`
+prefix:
+
+```
+/p2p/LongAhphaNumeric...
+```
+
+Look at your history of Data Credit consumption for the server, and forecast
+your target amount on that trend.  This example uses a nice round `1.0`
+billion DC.
+
+Convert HNT to DC via prevailing exchange rate from the
+[price oracle](https://api.helium.io/v1/oracle/prices/current).
+
+Optionally, get
+[predicted HNT Oracle Prices](https://api.helium.io/v1/oracle/predictions).
+"If no predictions are returned, the current HNT Oracle Price is valid for at
+least 1 hour."
+
+Calculate `$target_dc / $oracle_price / 1000` or as Lisp expression for
+target of 1.0 billion DC with HNT price of 2425000000 resulting in 412:
+
+```lisp
+(let ((current-hnt-price (/ 2425000000 1000))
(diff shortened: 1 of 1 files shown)


docs/runbook.md
```diff
@@ -0,0 +1,245 @@
...
+alias console_psql='docker exec -it helium_postgres psql -U postgres'
+alias console_restart='console_down && console_up'
+alias console_up='docker-compose build && docker-compose up -d && docker-compose logs -f'
+alias console_update='docker-compose build --pull && console_restart'
+```
+
+Miscellaneous:
+
+```bash
+alias alert='notify-send --urgency=low -i "$([ $? = 0 ] && echo terminal || echo error)" "$(history|tail -n1|sed -e '\''s/^\s*[0-9]\+\s*//;s/[;&|]\s*alert$//'\'')"'
```
michaeldjeffrey, line 40: I think this alias comes default on ec2 instances. We haven't used it for anything.

docs/runbook.md
```diff
@@ -0,0 +1,245 @@
...
+## Install A Release
+
+Published Docker images are
+[tagged](https://quay.io/repository/team-helium/router?tab=tags)
+as `production` or `staging` as appropriate.
+
+Install:
+
+```bash
+router install production
```
michaeldjeffrey, line 71: What does this do, I've never seen this command?

docs/runbook.md
```diff
@@ -0,0 +1,245 @@
...
+curl --head https://snapshots.helium.wtf/mainnet/snap-1279441.gz
+TZ=Z date
+```
+
+Scripted version:
+
+```bash
+wget -N \
+  $(curl https://snapshots.helium.wtf/mainnet/latest-snap.json | \
+    jq .height | \
```
michaeldjeffrey, line 184: Do any of our servers have `jq` installed?

docs/runbook.md
```diff
@@ -0,0 +1,245 @@
...
+rsync -a /tmp/latest.snap  192.168.123.123:/var/data/latest.snap
+```
+
+Once copied to the intended server, load that snapshot:
+
+```bash
+router snapshot load /var/data/latest.snap
+```
+
+## Tracing A Device
```
michaeldjeffrey, line 215: We might note that traces disable themselves after 240mins. 
Or if the server get's restarted.
If you need a device trace longer than that, the trace needs to be started in a different way.

docs/runbook.md
dpezely: Removed with commit 7e102c6

docs/runbook.md
dpezely: This bit of script is more for quick copy/paste from laptop/workstation in a pinch; e.g, 3am and half asleep.

That said: if such granularity seems excessive for this document, we can remove it.

docs/runbook.md
dpezely: Found this in a bit of bash history back-scroll but obviously haven't tried it.

Anyone else know?

docs/runbook.md
dpezely: What is the alternate way of starting such a trace?

docs/runbook.md
```diff
@@ -0,0 +1,266 @@
...
+with the filename being then-current block height.
+
+Similarly, locally generated snapshots should be within the directory path
+specified by `blockchain` `->` `base_dir`.
+
+### Manual Snapshots
+
+Start on a server confirmed to be current with respect to blockchain height.
+
+Maybe start on a validator, which uses the `minor` command rather than
```
jontow, line 194: s/minor/miner ?

docs/runbook.md
```diff
@@ -0,0 +1,266 @@
...
+`router` but otherwise works the same.
+
+```bash
+miner snapshot take latest.snap
+docker cp validator:/opt/miner/latest.snap /tmp/
+rsync -a /tmp/latest.snap  192.168.123.123:/var/data/latest.snap
```
jontow, line 200: worth adding '--progress' to rsync here for groggy people, or does this routinely complete in moments because local-to-local?

docs/runbook.md
```diff
@@ -0,0 +1,266 @@
...
+
+## Stop/Start
+
+On production servers, Console versus Router run on different hosts to
+isolate and distribute workloads.
+
+Because staging has far less traffic, everything runs on one host.
+
+```bash
+docker-compose down
```
jontow, line 52: worth noting here or somewhere early that docker-compose commands are meant to be run from specific directories?  it prints a fairly obvious error, but a short warning may speed recovery for unfamiliar folks

docs/runbook.md
```diff
@@ -0,0 +1,266 @@
...
+restarted.
+
+```bash
+router device trace --id=abc12345-bbbb-cccc-dddd-eeeeeeeeeeee
+tail -F /var/data/router/log/traces/abc12.log
+```
+
+## Erlang Expressions
+
+For one-off expressions **not** intended to impact the running BEAM, use `eval`:
```
jontow, line 245: I know of the chat history that influences the language choice here, but the context will be lost shortly:  suggest changing or augmenting "impact" to something like "persist inside" for clarity

docs/runbook.md
michaeldjeffrey: It's semi-hardcoded into the trace functionality right now. 
But there's the settings
```
application:get_env(router, device_trace_timeout, 240)
```
that can be changed before a trace of a different length.

docs/runbook.md
dpezely: Note to self: don't compose too much text when sleep-deprived.  

Fixed in b4cdf8e

docs/runbook.md
dpezely: Good call, as these files are just under 300 MB.  Fixed in 64b4652

docs/runbook.md
dpezely: Good catch!  Fixed in 64b4652

docs/runbook.md
dpezely: Updated in 64b4652 along with the comment from `#blockchain` regarding the env var.

docs/runbook.md
dpezely: Yes, that's better wording.  Fixed in b4aba82

michaeldjeffrey (dismissed):

docs/runbook.md
dpezely: Should we keep or remove the section on `router install ...`?
Is it actually used by anyone?

macpie: Personally never used it

dpezely: So what is the actual release procedure used?

My current understanding is that you build a new docker container, push that to Quay, tag it as "production" via Quay UI (is there an API for that?), and then simply bounce the container via the usual, `docker-compose down && docker-compose up -d`.

Yes?

dpezely: Since nobody seems to use this command, it's been removed from the runbook.

michaeldjeffrey (approved):

macpie (approved):

PR #686, opened 2022-04-07
Improve CTs

macpie:
This PR aims to remove the need for P2P swarm to be started for testing and instead use basic Erlang messaging.

It will use a `gen_server` to simulate devices and will also insert itself into an `ets table` to be read by the testing Console.

This should increase reliability of the test suites, execution time and also bring more simplicity.

### DIFF
--- a/test/router_test_device.erl
+++ b/test/router_test_device.erl
@@ -1,0 +1,386 @@
+%%%-------------------------------------------------------------------
+%% @doc
+%% == Router Test Device ==
+%% @end
+%%%-------------------------------------------------------------------
+-module(router_test_device).
+
+-behavior(gen_server).
+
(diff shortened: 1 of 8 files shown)


test/router_test_device.erl
```diff
@@ -0,0 +1,387 @@
...
+    #state{hotspot_name = Name, device = Device, device_app_key = AppKey} = State
+) ->
+    ok = router_test_console:insert_device(Device, AppKey),
+    lager:info("post init for hotspot ~p device ~p", [Name, router_device:id(Device)]),
+    {noreply, State};
+handle_continue(_Msg, State) ->
+    lager:warning("rcvd unknown continue msg: ~p", [_Msg]),
+    {noreply, State}.
+
+handle_call({join, Opts}, _From, State) ->
```
macpie, line 149: We should simulate offer here

PR #693, opened 2022-04-11
fix xor filter reporting error when not all filters are in use

michaeldjeffrey:
fixes error
```
RPC to 'router@127.0.0.1' failed: {'EXIT',
                                   {function_clause,
                                    [{lists,zip,
                                      [[{1,0},{2,0},{3,0},{4,0}],[]],
                                      [{file,"lists.erl"},{line,391}]},
                                     {lists,zip,2,
                                      [{file,"lists.erl"},{line,391}]},
                                     {router_cli_xor_filter,filter_report,3,
                                      [{file,
                                        "/opt/router/src/cli/router_cli_xor_filter.erl"},
                                       {line,221}]},
                                     {clique_command,run,1,
```

### DIFF
--- a/src/cli/router_cli_xor_filter.erl
+++ b/src/cli/router_cli_xor_filter.erl
@@ -208,9 +208,13 @@ filter_report(["filter", "report"], [], [{id, ID}]) ->
     report_device(DeviceID);
 filter_report(["filter", "report"], [], []) ->
     [
-        {routing, Routing},
-        {in_memory, Memory}
+        {routing, Routing0},
+        {in_memory, Memory0}
     ] = router_xor_filter_worker:report_filter_sizes(),
+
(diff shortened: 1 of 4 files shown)


dpezely (changes requested):
Please add a comment explaining the rationale behind the logic; 
e.g.,  more than just the fact that these lists need to be of a particular length-- but why that's significant.

michaeldjeffrey:
> Please add a comment explaining the rationale behind the logic; e.g., more than just the fact that these lists need to be of a particular length-- but why that's significant.

I thought about this for a bit, and ended up moving the sizing to the cli module.
For the XOR filter worker, the size of the list is not significant.

dpezely (approved):

macpie (approved):

Issue #695, opened 2022-04-12
Mechanism to allow sending MAC commands from LNS to End Node on Dev


illperipherals:
Enable Port 0 to be allowed for downlinks on Dev. This is necessary for sending MAC commands to end-nodes.
Initially, this is only required for the Dev testing server.

PR #696, opened 2022-04-12
Add options to enable downlinking on port 0

macpie:
Fix for https://github.com/helium/router/issues/695


src/device/router_device_channels_worker.erl
```diff
@@ -558,6 +558,11 @@ downlink_decode(MapPayload) when is_map(MapPayload) ->
                 case maps:find(<<"port">>, MapPayload) of
                     {ok, X} when is_integer(X), X > 0, X < 224 ->
                         X;
+                    0 ->
```
michaeldjeffrey, line 561:
```suggestion
                    {ok, 0} ->
```

dpezely (dismissed):
Thanks for renaming to `queue_downlink()` because this will make it easier to understand for new contributors.

dpezely (approved):

PR #703, opened 2022-04-20
Various fixes

macpie:


.env-template
```diff
@@ -1,46 +1,54 @@
...
-ROUTER_XOR_FILTER_WORKER=false
+# Max time to wait for uplinks in ms, only for disco mode device
+# Default: 8 seconds blocks when not set
+ROUTER_DISCO_FRAME_TIMEOUT=8000
 
 # Turn on/off HTTP integration url checks like DNS, IP... (anything else than false is on, enabled by default)
 ROUTER_HTTP_CHANNEL_URL_CHECK=true
 
-# Set max number fo downlink allowed to be queued for one device (Defaults to 20).
+# Set max number fo downlink allowed to be queued for a device (Defaults to 20).
```
dpezely, line 44: preexisting minor typo

```suggestion
# Set max number for downlink allowed to be queued for a device (Defaults to 20).
```

dpezely (dismissed):

src/metrics/router_metrics.erl
```diff
@@ -510,6 +511,21 @@ record_queues() ->
...
+-spec record_hotspot_reputations() -> ok.
+record_hotspot_reputations() ->
+    case router_hotspot_reputation:enabled() of
+        false ->
+            ok;
+        true ->
+            lists:foreach(
+                fun({Hotspot, Counter}) ->
+                    Name = blockchain_utils:addr2name(Hotspot),
+                    _ = prometheus_gauge:set(?METRICS_HOTSPOT_REPUTATION, [Name], Counter)
```
michaeldjeffrey, line 523: We're currently running 15k unique on prod.
Prometheus doesn't like over 20k values for labels.

macpie: This should not be that many hotspots, maybe we can set a limit?

src/cli/router_cli_hotspot_reputation.erl
```diff
@@ -80,6 +96,69 @@ hotspot_reputation_reset(["hotspot_reputation", "reset", B58], [], []) ->
...
+                        {blockchain_state_channel_v1:name(SC), 0};
+                    Actors ->
+                        {blockchain_state_channel_v1:name(SC), TotalDcs / Actors}
+                end
+            end,
+            ActiveSCs
+        )
+    ),
+    HotspotList = lists:concat(
+        lists:map(
```
dpezely, line 118: No need to change anything.  Just calling this out as potential threshold for when to introduce intermediate variables for params rather than inlined values; whereas, the previous use of `lists:maps()` a few lines above is short enough to keep as-is.

Mind you, I have lots of old code (especially in Common Lisp) where I didn't introduce enough intermediate vars, and after revisiting that code years or months later is why I'm mentioning this now.

Thoughts?  we can discuss in #router channel.

dpezely (approved):

src/cli/router_cli_hotspot_reputation.erl
```diff
@@ -35,16 +35,32 @@ hotspot_reputation_usage() ->
...
+            "hotspot_reputation reset <b58_hotspot_id>       - Reset hotspot's reputation to 0\n",
+            "hotspot_reputation sc [--over 10] [--rep 10]    - Display hotspots in state channel going over X time the average (Default 10)\n"
         ]
     ].
 
 hotspot_reputation_cmd() ->
     [
         [["hotspot_reputation", '*'], [], [], fun hotspot_reputation_get/3],
-        [["hotspot_reputation", "reset", '*'], [], [], fun hotspot_reputation_reset/3]
+        [["hotspot_reputation", "reset", '*'], [], [], fun hotspot_reputation_reset/3],
```
michaeldjeffrey, line 48: Can we rename this to just `rep` so we don't actualyl have to laern to spell?

src/router_hotspot_reputation.erl
```diff
@@ -69,7 +69,6 @@ track_offer(Offer) ->
             ok;
```
michaeldjeffrey, line 69: Can we still track offers when the system is disabled?
That way you can get a sense of things without denying gateways.

dpezely (approved):

michaeldjeffrey (approved):

PR #705, opened 2022-04-22
Correct maximum downlink sizes in AU915, AS923 and EU868

jamesl-dm:
The sizes listed for AU915 and AS923 are the MACPayload sizes
(8 bytes too large), instead of the max application payloads.

The size listed in EU868 (222) is the maximum size when repeaters
are used. All of the other regions correctly list the maximum size when
repeaters are not used.

* Reduce all AU915 and AS923 values by 8

* Correct the EU868 max downlink size from 222 -> 242

### DIFF
--- a/src/lora/lorawan_mac_region.erl
+++ b/src/lora/lorawan_mac_region.erl
@@ -40,19 +40,19 @@
 
 -define(US915_MAX_DOWNLINK_SIZE, 242).
 -define(CN470_MAX_DOWNLINK_SIZE, 242).
--define(AS923_MAX_DOWNLINK_SIZE, 250).
--define(AU915_MAX_DOWNLINK_SIZE, 250).
--define(EU868_MAX_DOWNLINK_SIZE, 222).
+-define(AS923_MAX_DOWNLINK_SIZE, 242).
+-define(AU915_MAX_DOWNLINK_SIZE, 242).
+-define(EU868_MAX_DOWNLINK_SIZE, 242).
 
 -define(AS923_PAYLOAD_SIZE_MAP, #{
-    0 => 59,
-    1 => 59,
-    2 => 59,
-    3 => 123,
-    4 => 250,
-    5 => 250,
-    6 => 250,
-    7 => 250
+    0 => 51,
+    1 => 51,
+    2 => 51,
+    3 => 115,
+    4 => 242,
+    5 => 242,
+    6 => 242,
+    7 => 242
 }).
 
 -define(CN470_PAYLOAD_SIZE_MAP, #{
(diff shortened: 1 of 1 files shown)


mikev:
This describes the issue correctly, but the proposed fix is not correct.

First, some background context: repeaters would communicate with the network (gateways) using only data rates that supported the maximum allowed MAC Payload Size of 250 bytes.  Thus, repeater data rates show a maximum payload size which is 20 bytes fewer when describing “Repeater Compatible” operation.

Second in the presence of the optional FOpts the application maximum payload should be reduced in size by 8 bytes.

The code MUST list the maximum packets size without FOpts and without the presence of Repeaters.  This allows us to correctly compute other sizes through subtraction.

Size with optional FOpts = Max - 8
Size for Repeater = Max - 20
Size for Repeater with optional FOpts = Max - 20 - 8

src/lora/lorawan_mac_region.erl
```diff
@@ -40,19 +40,19 @@
 
 -define(US915_MAX_DOWNLINK_SIZE, 242).
 -define(CN470_MAX_DOWNLINK_SIZE, 242).
```
mikev, line 42: Downlink sizes should be computed, the same as uplink sizes.

src/lora/lorawan_mac_region.erl
```diff
@@ -105,10 +105,10 @@
     1 => 51,
     2 => 51,
     3 => 115,
-    4 => 222,
```
mikev, line 108: Maximum repeater-compatible size is 230.

src/lora/lorawan_mac_region.erl
```diff
@@ -82,20 +82,20 @@
 }).
 
 -define(AU915_PAYLOAD_SIZE_MAP, #{
-    0 => 59,
```
mikev, line 85: AU915 and US915 Uplink maximums are similar, just offset by 2.

jamesl-dm: My understanding is that the maximums for the US915 payload (MACPayload - 8) are:
**11, 53, 125, 242,** 242, x, x, x, 53, 129, 242, 242, 242, 242
Whereas the corresponding maximums in AU915 are:
**51, 51, 51, 115,** 242, 242, 242, x, 53, 129, 242, 242, 242, 242 (when the uplink dwell time is off)
And:
x,  x, 11, 53, 125, 242, 242, x, 53, 129, 242, 242, 242, 242 (when the uplink dwell time is on)

I assumed from the code that you weren't handling the uplink dwell time.
It isn't relevant to the maximum downlink sizes, so it's probably best to list the larger values.

jamesl-dm:
Hi @mikev

It sounds from your comments like you want the maximum MACPayload sizes listed here, not the maximum application payload sizes? I assumed you need the application payload sizes because the majority of the existing values match the application payload size.  Also, when you try to queue an oversize uplink, it succeeds if the application payload size is <= the listed value. So if the listed values are supposed to be MACPayload sizes, the size check is incorrect.

So it sounds like you are planning on refactoring this at the same time as issue #706.
Do you want me to close this pull request?

macpie:
@jamesl-dm  is it safe to close this so you can reopen on the other side?

jamesl-dm:
Yes no problem

PR #732, opened 2022-05-11
Add region field support for downlinks

hwahby:
Add support for region field in queued downlinks sent from the console (https://github.com/helium/console/pull/1141)

### DIFF
--- a/include/router_device_worker.hrl
+++ b/include/router_device_worker.hrl
@@ -9,7 +9,7 @@
     'AS923_4'
 ]).
 
--type 'downlink.region'() ::
+-type downlink_region() ::
     'US915' | 'AU915' | 'EU868' | 'CN470' | 'AS923_1' | 'AS923_2' | 'AS923_3' | 'AS923_4'.
 
 -record(frame, {
@@ -75,5 +75,5 @@
     port :: non_neg_integer(),
     payload :: binary(),
     channel :: router_channel:channel(),
-    region :: 'downlink.region'()
+    region :: downlink_region()
 }).
--- a/include/router_device_worker.hrl
+++ b/include/router_device_worker.hrl
@@ -1,3 +1,17 @@
+-define(DOWNLINK_REGIONS, [
+    'US915',
+    'AU915',
+    'EU868',
+    'CN470',
+    'AS923_1',
+    'AS923_2',
+    'AS923_3',
+    'AS923_4'
+]).
+
+-type 'downlink.region'() ::
+    'US915' | 'AU915' | 'EU868' | 'CN470' | 'AS923_1' | 'AS923_2' | 'AS923_3' | 'AS923_4'.
+
 -record(frame, {
     mtype,
     devaddr,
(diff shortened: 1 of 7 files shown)


include/router_device_worker.hrl
```diff
@@ -1,3 +1,17 @@
...
+    'AU915',
+    'EU868',
+    'CN470',
+    'AS923_1',
+    'AS923_2',
+    'AS923_3',
+    'AS923_4'
+]).
+
+-type 'downlink.region'() ::
```
macpie, line 12:
```suggestion
-type downlink_region() ::
```
Do you mind if we do this instead? We have an aversion against `'` on atoms

src/device/router_device_channels_worker.erl
```diff
@@ -544,7 +546,7 @@ maybe_report_downlink_dropped(DeviceID, Desc, Channel) ->
     end.
 
 -spec downlink_decode(binary() | map()) ->
-    {ok, {boolean(), integer(), binary()} | clear_queue} | {error, any()}.
+    {ok, {boolean(), integer(), atom(), binary()} | clear_queue} | {error, any()}.
```
macpie, line 549: This could be your region type here instead of `atom()`
```suggestion
    {ok, {boolean(), integer(), downlink_region(), binary()} | clear_queue} | {error, any()}.
```

mikev:
We may need to rethink this PR.  The concept of 'Region' will go away soon. 

To understand why: we can somewhat think of Regions in an OOP sense: the Region is the class and the Plan is the instance.  Currently we have a 1-1 correspondence between class and instance, but this is now going away.  Malaysia is a specific example, but there will be others soon.   With Malaysia we have defined a new Region named 'AS923_1B'.  This is fully an AS923 plan but we've defined channel frequencies which don't use 924Mhz which is illegal in Malaysia.  

My concern with this PR is that if we hard-code these Region strings into Router, we will shortly need to sort how to handle this with the new Malaysia plan which is AS923_1B.  It also turns out that we cannot reuse Malaysia for Australia, so we'll soon have a new plan named AS923_1C.

macpie:
> We may need to rethink this PR. The concept of 'Region' will go away soon.
> 
> To understand why: we can somewhat think of Regions in an OOP sense: the Region is the class and the Plan is the instance. Currently we have a 1-1 correspondence between class and instance, but this is now going away. Malaysia is a specific example, but there will be others soon. With Malaysia we have defined a new Region named 'AS923_1B'. This is fully an AS923 plan but we've defined channel frequencies which don't use 924Mhz which is illegal in Malaysia.
> 
> My concern with this PR is that if we hard-code these Region strings into Router, we will shortly need to sort how to handle this with the new Malaysia plan which is AS923_1B. It also turns out that we cannot reuse Malaysia for Australia, so we'll soon have a new plan named AS923_1C.
> 
> Lets discuss...

I agree this will evolve and those regions should come from the loran lib but for the moment this allows us to queue downlinks before any join/uplink are sent. This is what this PR is really about.

include/router_device_worker.hrl
hwahby: Addressed.

src/device/router_device_channels_worker.erl
hwahby: Addressed.

macpie (approved):

PR #733, opened 2022-05-11
Fix join devaddr bug 

macpie:
This PR includes 

- Update reputation system to include `UnknownDevice` delivered packets.
- Make `devaddr` field a list (`devaddrs`) in `router_device` record so that Router can track  multiple devaddrs. until device actually send first `uplink`

### DIFF
--- a/include/router_device.hrl
+++ b/include/router_device.hrl
@@ -1,3 +1,25 @@
+-record(device_v7, {
+    id :: binary() | undefined,
+    name :: binary() | undefined,
+    dev_eui :: binary() | undefined,
+    app_eui :: binary() | undefined,
+    %% {nwk_s_key, app_s_key}
+    keys = [] :: list({binary() | undefined, binary() | undefined}),
+    devaddrs = [] :: [binary()],
+    dev_nonces = [] :: [binary()],
+    fcnt = 0 :: non_neg_integer(),
+    fcntdown = 0 :: non_neg_integer(),
+    offset = 0 :: non_neg_integer(),
+    channel_correction = false :: boolean(),
+    queue = [] :: [any()],
+    region :: atom() | undefined,
+    last_known_datarate :: integer() | undefined,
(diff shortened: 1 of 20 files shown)


include/router_device.hrl
```diff
@@ -1,3 +1,25 @@
+-record(device_v7, {
```
dpezely, line 1: This is a good opportunity to add `rx_delay` to device record, also.

Remember that `rx_delay` has multiple components: 
1. `metadata` contains user-supplied value from Console API
2. but we also *abuse* `metadata` for tracking state: 
    + Previously established value
    + plus whether request has been sent via downlink but not yet ACK'd by device

src/device/router_device_cache.erl
```diff
@@ -52,13 +59,24 @@ get(DeviceID) ->
...
+        [] -> [];
+        List -> [Device || {_DevAddr, Device} <- List]
+    end.
 
 -spec save(router_device:device()) -> {ok, router_device:device()}.
 save(Device) ->
     DeviceID = router_device:id(Device),
     true = ets:insert(?ETS, {DeviceID, Device}),
+    % MS = ets:fun2ms(fun({_, D}) when D#device_v7.id == DeviceID -> true end),
+    MS = [{{'_', '$1'}, [{'==', {element, 2, '$1'}, {const, DeviceID}}], [true]}],
```
dpezely, line 72: What are the semantics of `MS` in this context?  Nothing was apparent from sibling functions.  Please add a code comment.

dpezely (dismissed):
We know that this is robust since we abused `router-dev` running this code when receiving benchmark runs that caused devaddr collisions.

The `rx_delay` concern can be another PR but should be addressed or punted before release.

dpezely (dismissed):

dpezely (dismissed):

dpezely (approved):

PR #765, opened 2022-05-25
Light Router: Connect to validator to get chain vars

macpie:
#721


macpie:
This will be used as an example for futur use.

PR #777, opened 2022-06-01
Various fixes

macpie:


src/router_hotspot_reputation.erl
```diff
@@ -82,10 +82,25 @@ track_packet(SCPacket) ->
...
-    lager:info("hotspot ~p unknown_device= ~p", [blockchain_utils:addr2name(Hotspot), Counter]),
+-spec track_unknown_device(
+    Packet :: blockchain_helium_packet_v1:packet(), Hotspot :: libp2p_crypto:pubkey_bin()
+) -> ok.
+track_unknown_device(Packet, Hotspot) ->
+    erlang:spawn(fun() ->
+        PHash = blockchain_helium_packet_v1:packet_hash(Packet),
+        % MS = ets:fun2ms(fun({Key, _Time}) when Key =:= {Hotspot, PHash} -> true end)
+        MS = [{{'$1', '$2'}, [{'=:=', '$1', {{Hotspot, PHash}}}], [true]}],
+        case ets:select_delete(?OFFER_ETS, MS) of
```
macpie, line 93: change this to normal lookup and delete

src/apis/router_console_dc_tracker.erl
```diff
@@ -180,6 +180,8 @@ handle_info(
     {blockchain_event, {add_block, _BlockHash, _Syncing, _Ledger}},
     #state{chain = undefined} = State
 ) ->
+    lager:info("got block ~p with not chain", [_BlockHash]),
+    erlang:send_after(500, self(), post_init),
```
michaeldjeffrey, line 184: We add a post_init message in `init`. 
Isn't this just going to add another check every 500ms for every block we receive without a chain?
I think it's highly unlikely, but also a little unnecessary.

dpezely (approved):
Approved, but wait for at least one other to confirm since I was under the weather this week.

PR #778, opened 2022-06-01
don't crash on unavailable rx2 windows

michaeldjeffrey:
unknown regions come through and crash the worker, which needs to startup for every gateway reporting. Causing a bad loop.


macpie:
This was just in case, closing

Issue #643, opened 2022-03-05
Fcnt rollover not correctly handled


mrx23dot:
Last valid packet up:
![2022-03-05_114720_IB](https://user-images.githubusercontent.com/4163396/156881668-aae78ff8-578e-425c-b84a-803796bfdc83.jpg)

Since Fcnt is 16bit in frame, the next valid counter is 0, but it gets dropped.
Sending Fcnt=1 is dropped for the same reason.
Sending 65534, 65535 again is obviously dropped.

So my conclusion rollover is not handled on Helium.
It's easy to implement  for v1.0.4
`u16 diff = u16 new - u16 last // handles rollover`  
diff > 0

```
[
  {
    "category": "uplink_dropped",
    "data": {
      "fcnt": 65535,
      "hold_time": 623,
      "hotspot": {
        "id": "112K3K82t8VagR8kVmrPGqcQKftaxp2P9pMpaDKY7mDXwt5ToXpU",
        "name": "real-shadow-gecko"
      }
    },
    "description": "Late packet",
    "device_id": "66e67e40-90ff-4177-a23b-29c33f4d6c35",
    "frame_down": null,
    "frame_up": 65535,
    "id": "e4d7e1fb-f63a-4bc1-adfc-17974e8a7383",
    "organization_id": "32595cfd-96a9-4510-96dd-826b546711db",
    "reported_at": "1646480468954",
    "router_uuid": "829ae5ba-7690-47b5-ad4e-75fde0fd764d",
    "sub_category": "uplink_dropped_late"
  },
  
  -----------------------------------------
  {
    "category": "uplink",
    "data": {
      "integration": {
        "id": "no_channel",
        "name": "no_channel",
        "status": "no_channel"
      },
      "req": {}
    },
    "description": "No Channel Configured",
    "device_id": "66e67e40-90ff-4177-a23b-29c33f4d6c35",
    "frame_down": null,
    "frame_up": null,
    "id": "94ba9b82-6a53-44b3-b986-5c8935cc62e5",
    "organization_id": "32595cfd-96a9-4510-96dd-826b546711db",
    "reported_at": "1646480426775",
    "router_uuid": "2ec02a8b-ddfb-409a-8f8c-6f8dc7c1f8af",
    "sub_category": "uplink_integration_req"
  },
  {
    "category": "uplink",
    "data": {
      "dc": {
        "balance": 8711,
        "nonce": 1,
        "used": 1
      },
      "devaddr": "44000048",
      "fcnt": 65535,
[...]

PR #780, opened 2022-06-03
[Issue #643] Implment the uplink frame counter as a 32-bit integer

barndon:
This PR addresses issue #643.  In a nutshell, it updates Router to correctly implement the uplink frame counter as a 32-bit integer.  The details are as follows:

1.  The default value of `fcnt` in a device record is now `undefined`.  (Within the code, the device's `fcnt` value should be understood to mean "The last successfully processed FCnt value.")
2.  `router_device:fcnt_next_val/1` has been added.  This function returns an integer indicating what FCnt value the next valid packet should have.  If the device's `fcnt` is `undefined`, the next value is 0.
3.  `router_device_routing` is updated verify the packet's message integrity code (MIC) using the 32-bit frame counter value instead of the 16-bit fcnt value obtained from the packet.
4. Our method of arriving at the 32-bit frame counter value is as follows: 

  - If this is the first frame after a successful join, we assume the high word is 0 and we take the lower word as-is from the FCnt field of the packet.
  - If this is not the first packet, then we have a frame counter value from the previous packet; if the FCnt field from the packet is greater than or equal to the low word of our frame counter, we join the high word of our counter to the FCnt field from the packet and accept that as the new frame counter value.
  - If this is not the first packet, then we have a frame counter from the previous packet; if the FCnt field from the packet is "a little bit less" than the low word of our counter (meaning the difference is =< 10), we join the high word of our counter to the FCnt field from the packet and accept that as the new frame counter value.
  - If this is not the first packet, then we have a frame counter from the previous packet; if the FCnt is "a lot less" than the low word of our frame counter (meaning the difference is more than 10), we increment the high word of our counter by 1, join that to the FCnt field from the packet, and accept the result as the new frame counter value.

6.  `router_device_worker` is updated to accept a frame counter value from routing; it no longer considers the fcnt value from the packet.  When a device successfully joins, the worker sets the device's fcnt field to `undefined`.

A similar fix is also applied to the downlink frame counter.  The details are slightly different:

1. The default value of `fcntdown` in a device record remains 0.
2. `router_device:fcntdown_next_val/1` has been added.  This function returns an integer indicating what the value of `fcntdown` should be for the next downlink.  Under this function, the "next" value after 16#FFFFFFFF is 0.
3. `router_device_worker` is updated to use `fcntdown_next_val/1` to increment `fcntdown`.


test/test_utils.erl
```diff
@@ -803,7 +803,7 @@ frame_packet(MType, PubKeyBin, NwkSessionKey, AppSessionKey, FCnt, Options) ->
         type = lorawan,
         payload = Payload1,
         frequency = 923.3,
-        datarate = maps:get(datarate, Options, <<"SF8BW125">>),
+        datarate = maps:get(datarate, Options, "SF8BW125"),
```
mikev, line 806: FYI - The Erlang LoRaWAN library assumes data rates are always of the form <<"SF8BW125">>.  It could handle the second form you changed it to, but not yet...

test/test_utils.erl
```diff
@@ -840,10 +840,10 @@ frame_payload(MType, DevAddr, NwkSessionKey, AppSessionKey, FCnt, Options) ->
     Data = lorawan_utils:reverse(
         lorawan_utils:cipher(Body, AppSessionKey, MType band 1, DevAddr, FCnt)
     ),
-    FCntSize = maps:get(fcnt_size, Options, 16),
+    <<FCntLow:16/integer-unsigned-little, _:16>> = <<FCnt:32/integer-unsigned-little>>,
     Payload0 =
         <<MType:3, MHDRRFU:3, Major:2, DevAddr:4/binary, ADR:1, ADRACKReq:1, ACK:1, RFU:1,
-            FOptsLen:4, FCnt:FCntSize/little-unsigned-integer, FOptsBin:FOptsLen/binary,
+            FOptsLen:4, FCntLow:16/little-unsigned-integer, FOptsBin:FOptsLen/binary,
```
mikev, line 846: Good catch!  Yeah, the payload's fcnt is always 16 bits, never 32.

src/device/router_device.erl
```diff
@@ -58,6 +59,7 @@
 ]).
 
 -define(QUEUE_SIZE_LIMIT, 20).
+-define(MAX_32_BITS, 16#100000000).
```
michaeldjeffrey, line 62: This comes out to 33bits.

There's a util for this, `router_device_devaddr:uint32/1`, might be a good opportunity to move it to `router_utils`.

src/device/router_device_routing.erl
```diff
@@ -913,36 +919,44 @@ send_to_device_worker_(Packet, PacketTime, HoldTime, Pid, PubKeyBin, Region, Dev
...
                         Packet,
                         PacketTime,
                         HoldTime,
                         PubKeyBin,
                         Region,
                         Pid
-                    )
+                    ),
+                    Device1 = router_device:update([{fcnt, FCnt}], Device),
+                    ok = save_device(Device1),
```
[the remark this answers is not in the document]
barndon: Yes!  Done.

dpezely (approved):
Looks great especially since all of the comments from others seem to be addressed.

Again, for naming the operand for `rem` my suggestion was "MODULO_32_BITS" which is buried within the resolved discussion thread.  Since it's not exported, it probably doesn't matter.

include/router_device.hrl
```diff
@@ -126,7 +126,7 @@
     nwk_s_key :: binary() | undefined,
     app_s_key :: binary() | undefined,
     join_nonce = <<>> :: binary(),
-    fcnt = 0 :: non_neg_integer(),
+    fcnt = undefined :: undefined | non_neg_integer(),
```
mikev, line 129: Why do we need to update device_v1?
Should we be creating a device_v8?

src/device/router_device_worker.erl
```diff
@@ -1387,7 +1402,7 @@ validate_frame(
             validate_frame_(
                 PacketFCnt, Packet, PubKeyBin, Region, Device0, OfferCache, Blockchain, false
             );
-        undefined when FrameAck == 0 andalso PacketFCnt =< LastSeenFCnt ->
+        undefined when FrameAck == 0 andalso PacketFCnt =:= LastSeenFCnt ->
```
macpie, line 1405: I don't know if this is right, what if a hotspot sent delivered a "very old" packet?

test/test_utils.erl
michaeldjeffrey: protobufs always expect strings for datarate.

src/device/router_device_worker.erl
barndon: Resolved by the latest series of commits.

dpezely (dismissed):
There's orphaned code in a comment starting Line 1197 of
`src/devices/router_device_routing.erl` but otherwise looks good.

dpezely (approved):

Issue #745, opened 2022-05-17
Use device keys from console instead of router self generating device keys


vicmgs:
Currently router self generates device keys for use with aws. We will use console generated keys going forward.

related: https://github.com/helium/console/issues/1088

macpie:
See https://github.com/helium/router/pull/789#issuecomment-1151709847

Migration will be done in here as well

PR #789, opened 2022-06-08
Receive ecc device keys from Console

hwahby:
Will move to Console generating keys needed to interact with AWS, instead of generating on Router.

https://github.com/helium/console/issues/1088
https://github.com/helium/router/issues/745

### DIFF
--- a/src/device/router_device_worker.erl
+++ b/src/device/router_device_worker.erl
@@ -374,18 +374,28 @@ handle_cast(
                         []
                 end,
 
-            DeviceUpdates = [
-                {name, router_device:name(APIDevice)},
-                {dev_eui, router_device:dev_eui(APIDevice)},
-                {app_eui, router_device:app_eui(APIDevice)},
-                {devaddrs, DevAddrs},
-                {metadata,
-                    maps:merge(
-                        lorawan_rxdelay:maybe_update(APIDevice, Device0),
-                        router_device:metadata(APIDevice)
-                    )},
-                {is_active, IsActive}
-            ],
+            ECCUpdate =
+                case router_device:ecc_compact(APIDevice) of
+                    undefined ->
+                        [];
+                    ECC ->
+                        DecodeECC = base64:decode(ECC),
+                        [{ecc_compact, libp2p_crypto:bin_to_pubkey(DecodeECC)}]
+                end,
+
+            DeviceUpdates =
+                [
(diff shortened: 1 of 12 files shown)


src/device/router_device_worker.erl
```diff
@@ -374,18 +374,28 @@ handle_cast(
...
-                        lorawan_rxdelay:maybe_update(APIDevice, Device0),
-                        router_device:metadata(APIDevice)
-                    )},
-                {is_active, IsActive}
-            ],
+            ECCUpdate =
+                case router_device:ecc_compact(APIDevice) of
+                    undefined ->
+                        [];
+                    ECC ->
```
michaeldjeffrey, line 381: From the device record.
`ecc_compact :: map() | undefined,`

I think that's going to break `base64:decode/1`

src/device/router_device_worker.erl
```diff
@@ -374,18 +374,28 @@ handle_cast(
...
+                    {dev_eui, router_device:dev_eui(APIDevice)},
+                    {app_eui, router_device:app_eui(APIDevice)},
+                    {devaddrs, DevAddrs},
+                    {metadata,
+                        maps:merge(
+                            lorawan_rxdelay:maybe_update(APIDevice, Device0),
+                            router_device:metadata(APIDevice)
+                        )},
+                    {is_active, IsActive}
+                ] ++ ECCUpdate,
```
michaeldjeffrey, line 398: We might need to move this to the api worker so it comes in with every API device.

macpie:
We will be doing the migration in here as well. 

`router_device_worker:get_device(DB, CF, DeviceID)` (the one used in `init/1`) should check for keys from `Device1` and compare with keys from disk (`Device0`) if they are different Router should ignore Console and also make an API call to Console to update the keys

src/device/router_device_worker.erl
```diff
@@ -374,18 +374,28 @@ handle_cast(
...
-                    )},
-                {is_active, IsActive}
-            ],
+            ECCUpdate =
+                case router_device:ecc_compact(APIDevice) of
+                    undefined ->
+                        [];
+                    ECC ->
+                        DecodeECC = base64:decode(ECC),
+                        [{ecc_compact, libp2p_crypto:bin_to_pubkey(DecodeECC)}]
```
macpie, line 383: here you will need to use `libp2p_crypto:keys_from_bin/1`

PR #790, opened 2022-06-09
Support AS923-AU915 dual-plan

mikev:


dpezely (approved):

macpie (approved):

src/device/router_device_worker.erl
```diff
@@ -1464,12 +1501,18 @@ validate_frame_(Packet, PubKeyBin, HotspotRegion, Device0, OfferCache, Blockchai
                             AName
                         ]
                     ),
-                    {Region, DR} = packet_datarate(Device0, Packet, HotspotRegion),
+                    Region = region_or_default(router_device:region(Device0), HotspotRegion),
```
michaeldjeffrey, line 1504: Can we do all of this in the `packet_datarate/3`?

mikev: packet_datarate is DEAD_CODE now.

michaeldjeffrey: All 3 calls to `packet_datarate/3` were replaced with the same snippet that give us exactly what `packet_datarate/3` gave us.
And if it's truly dead code, we can remove it.

PR #800, opened 2022-06-21
Add Throttling to routing

macpie:


src/device/router_device_routing.erl
```diff
@@ -100,6 +105,10 @@ init() ->
         ?BF_ROTATE_AFTER
     ),
     true = ets:insert(?BF_ETS, {?BF_KEY, BloomJoinRef}),
+    HotspotRateLimit = application:get_env(router, hotspot_rate_limit, 10),
+    DeviceRateLimit = application:get_env(router, device_rate_limit, 1),
```
dpezely, line 109: What are the units for this value?  e.g., N *seconds between requests*? N requests per second?

dpezely (commented):
~~The `throttle` package on hex.pm requires redis, so where does that get configured?~~

[EDIT: never mind.  missed `lambda_throttle` being aliased to "throttle"]

dpezely (approved):

PR #840, opened 2022-08-01
pass in pid of the grpc handler to state channels server

andymck:
fix incorrect PID being passed to blockchain_state_channels_server:track_offer/3

Correct PID is that of the handler ( in this case the grpc handler )

This will have prevented correct responses being returned to the client as the grpc handler will never receive the response itself and subsequently time out and return the default "no response" msg


macpie:
We actually don't care about the return of the offer anymore as we just account for it after.

PR #850, opened 2022-08-11
Upgrade validator to validator1.13.7

macpie:
This is being tested. An tang was created `2022.08.10.0` to test on staging as well


macpie:
@michaeldjeffrey I found a good one here https://github.com/helium/router/pull/850/commits/f6e15660e1aff65d4b37e4ec15665b08538cd128

The issue with a blank select delete is we remove (temporarily) all `devaddr` from that table if unlucky (happening frequently with my device) a packet can show up from a hotspot and not get routed as the device cache is saving.

Since patch it seems that Router is not missing any uplinks anymore from my 3 hotspots, see below:

Screen Shot 2022-08-10 at 7 48 57 PM

michaeldjeffrey (approved):

PR #853, opened 2022-08-25
Expand instructions for Load Snapshot

dpezely:
This only touches the `docs/` subdirectory to update the runbook.

For when on-call, please proofread with perspective of being sleepy at 3AM.

- True story based upon recent events
- Names and locations have been changed to protect the innocent

### DIFF
--- a/docs/runbook.md
+++ b/docs/runbook.md
@@ -208,10 +208,70 @@ docker cp validator:/opt/miner/latest.snap /tmp/
 rsync -a --progress /tmp/latest.snap  192.168.123.123:/var/data/latest.snap
 ```
 
+### Load Snapshot
+
 Once copied to the intended server, load that snapshot:
 
+Ensure that the snapshot file lives under `/var/data/`.  Otherwise, the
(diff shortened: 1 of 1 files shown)


docs/runbook.md
```diff
@@ -208,12 +208,72 @@ docker cp validator:/opt/miner/latest.snap /tmp/
...
+tail -F /var/data/log/router.log
+```
+
+Note: uppercase `-F` flag, in case log files get rotated.
+
+When a log message appears mentioning `maybe_start_state_channel`, restart
+the container **again**.
+
+In the meantime, running `router sc list` will return `none`.  Ignore it.
+Loading probably is continuing in the background.
```
michaeldjeffrey, line 265:
```suggestion
Loading probably is continuing in the background.

There's a chance the snap loaded contains closing txns for all the open State Channels. In that case, Router will need to open new State Channels. You can view those with `router txn queue`.
```

mikev (approved):

michaeldjeffrey (approved):

PR #861, opened 2022-09-13
add grpc service for hpr packets

michaeldjeffrey:
- allows us to keep signatures intact from gateways while routing

Made a new grpc client config as to not touch things that are already working. This client is a little different in that the fact that it uses the proto records rather than maps. Otherwise, basically the same.


michaeldjeffrey:
I have this loaded up on dev right now, and have had a successful test run with a virtual-device and gateway-rs.

macpie (approved):

PR #862, opened 2022-09-29
update erlang_lorawan ref to latest

mikev:


marcsugiyama:
The github workflow for `xref` doesn't generate the grpc files in `src/grpc/autogen`, and these files don't appear to be included in the build artifacts.

Makefile
```diff
@@ -74,8 +74,10 @@ grpc:
 $(grpc_services_directory):
 	@echo "grpc service directory $(directory) does not exist, generating services"
 	$(REBAR) get-deps
-	$(MAKE) grpc
+	REBAR_CONFIG="config/grpc_server_gen.config" $(REBAR) grpc gen
```
marcsugiyama, line 77: Could this be `make grpc` instead to avoid duplicating the generation steps?

michaeldjeffrey: I tried that and it fell into a loop.

macpie (approved):

PR #903, opened 2023-02-01
SKF

macpie:


src/device/router_device_worker.erl
```diff
@@ -733,18 +732,45 @@ handle_cast(
...
                     {dev_nonces, DevNonces},
                     {devaddrs, [DevAddr1]},
                     {fcnt, PacketFCnt}
                 ],
                 D1 = router_device:update(DeviceUpdates, Device0),
                 ok = save_device(DB, CF, D1),
+
+                ToAdd = [{add, DevAddr0, UsedNwkSKey}],
+                %% We have to usort just in case DevAddr assigned is the same
+                ToRemove0 = lists:usort(
```
michaeldjeffrey, line 751: Some alternatives.
You can `lists:flatmap/2` on the outside, and `lists:map/2` on the inside to not have to worry about the `Acc`s.

Or somehow grab all the devaddrs as ints and do a list comprehension.
```erlang
[{remove, DevAddr, Key} || DevAddr <- DevAddrInts, Key <- Keys]
```

src/device/router_device_worker.erl
```diff
@@ -733,18 +732,45 @@ handle_cast(
...
+                                Acc0,
+                                router_device:devaddrs(Device0)
+                            )
+                        end,
+                        [],
+                        Keys
+                    )
+                ),
+                %% Making sure that the pair that was added is not getting removed (just in case DevAddr assigned is the same)
+                ToRemove1 = ToRemove0 -- [{remove, DevAddr0, UsedNwkSKey}],
```
michaeldjeffrey, line 770: Making sure I understand this comment. 
Wouldn't this only be a problem if both DevAddr and NwkSKey are the same?

Should we flip around adding and removing just in case?
```erlang
ToRemove1 ++ ToAdd
```

test/router_test_ics_route_service.erl
```diff
@@ -130,8 +130,8 @@ verify_list_req(Req) ->
         libp2p_crypto:bin_to_pubkey(Req#iot_config_route_list_req_v1_pb.signer)
     ).
 
--spec verify_update_euis_req_req(Req :: #iot_config_route_update_euis_req_v1_pb{}) -> boolean().
-verify_update_euis_req_req(Req) ->
+-spec verify_update_euis_req(Req :: #iot_config_route_update_euis_req_v1_pb{}) -> boolean().
+verify_update_euis_req(Req) ->
```
michaeldjeffrey, line 134: RIP old name_req 😞

src/device/router_device_worker.erl
michaeldjeffrey: I don't know if the order thing matters anymore, on the other side the proplist is turned into a map.

michaeldjeffrey (approved):

PR #944, opened 2023-03-16
thread the found device through handling packet

michaeldjeffrey:
When we're handling grpc packets, we know the correct device up front, if we have the device, use it all the way through.

If we don't have the device, do the normal finding of the device for the packet.

### DIFF
--- a/src/device/router_device_routing.erl
+++ b/src/device/router_device_routing.erl
@@ -174,7 +174,7 @@ handle_packet(SCPacket, PacketTime, Pid) when is_pid(Pid) ->
     Packet = blockchain_state_channel_packet_v1:packet(SCPacket),
     Region = blockchain_state_channel_packet_v1:region(SCPacket),
     HoldTime = blockchain_state_channel_packet_v1:hold_time(SCPacket),
-    case packet(Packet, PacketTime, HoldTime, PubKeyBin, Region, Pid) of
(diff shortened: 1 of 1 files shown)


src/device/router_device_routing.erl
```diff
@@ -863,6 +864,7 @@ lookup_bf(Key) ->
     Pid :: pid()
 ) -> ok | {error, any()}.
 packet(
+    _Device,
```
michaeldjeffrey, line 867: This clause is for joins, and we request the device from Console by EUI to make sure we have the most up to date version, we want to continue to do that. So we ignore the passed device here.

src/device/router_device_routing.erl
```diff
@@ -1037,24 +1043,47 @@ send_to_device_worker(
...
+                                DevAddr,
+                                lorawan_utils:binary_to_hex(DevAddr),
+                                blockchain_utils:addr2name(PubKeyBin)
+                            ]
+                        ),
+                        {error, unknown_device};
+                    {ok, {_Device, _NwkSKey, _FCnt} = DeviceInfo0} ->
+                        DeviceInfo0
+                end;
+            _Device ->
```
michaeldjeffrey, line 1064: When a device is passed in, we still need to get the correct NwkSKey and FCnt.

macpie (approved):

PR #949, opened 2023-03-20
add a configurable delay while sending large batches of updates

michaeldjeffrey:
If too many items are sent to the config service, it will get stuck reading everything from the network before it will continue doing any db operations, this can be very slow (think minutes). If we added a few strategic pauses to our streaming up of items, the config service can keep up and hopefully not fall behind, making the request seem to take only as long as it takes to send everything up.

### DIFF
--- a/src/grpc/router_ics_eui_worker.erl
+++ b/src/grpc/router_ics_eui_worker.erl
@@ -354,6 +354,9 @@ maybe_update_euis(List0, State) ->
 update_euis([], _State) ->
     ok;
 update_euis(List, State) ->
+    BatchSleep = router_utils:get_env_int(config_service_batch_sleep_ms, 500),
+    MaxAttempt = router_utils:get_env_int(config_service_max_timeout_attempt, 5),
+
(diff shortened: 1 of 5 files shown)


src/grpc/router_ics_eui_worker.erl
```diff
@@ -362,35 +365,29 @@ update_euis(List, State) ->
...
             ),
+
             ok = grpcbox_client:close_send(Stream),
-            wait_for_stream_close(
-                init,
-                Stream,
-                0,
-                router_utils:get_env_int(eui_max_timeout_attempt, 5)
-            )
+            lager:info("done sending eui update ltimeout_retry: ~p]", [MaxAttempt]),
```
macpie, line 389:
```suggestion
            lager:info("done sending eui updates timeout_retry: ~p]", [MaxAttempt]),
```

macpie (dismissed):

macpie:
maybe add a quick `eunit` for this?

PR #952, opened 2023-03-22
skf reconcile fix

michaeldjeffrey:
There was a mismatch in `session_key` types. proto and the service was using lists. Our local construction for comparing what we already have was using binaries. So any time we tried to reconcile, it would try to remove everything, and re-add everything in the same update.


macpie (dismissed):

config/sys.config.src
```diff
@@ -56,8 +56,10 @@
             devaddr_enabled => "${ROUTER_ICS_DEVADDR_ENABLED}",
             route_id => "${ROUTER_ICS_ROUTE_ID}"
         }},
-        {config_service_max_timeout_attempt, 5},
+        {config_service_max_timeout_attempt, 25},
+        {config_service_max_timeout_sleep_ms, 5000},
         {config_service_batch_sleep_ms, 500},
+        {config_service_batch_size, 500},
```
michaeldjeffrey, line 62: This will send 500 SKF updates, waiting 0.5s in between, in hopes of keeping allowing the config service to not fall behind the stream. Deletes take a longer time than Additions, so we will wait for up to 2 minutes waiting for the stream to close properly.

rebar.config
```diff
@@ -27,6 +27,7 @@
     {elli, "3.3.0"},
     {bbmustache, {git, "https://github.com/helium/bbmustache", {branch, "data-fun"}}},
     {iso8601, {git, "https://github.com/erlsci/iso8601", {branch, "master"}}},
+    {chatterbox, {git, "https://github.com/tsloughter/chatterbox", {branch, "master"}}},
```
michaeldjeffrey, line 30: Going to upstream chatterbox for grpcbox improved skf list performance by more than 10x.
If a Router has a significant amount of devices, this can have a huge impact. 
Listing SKFs at the start of a reconcile can take up to 2 minutes for a Router with more than 75k devices under best conditions.

src/grpc/router_ics_eui_worker.erl
```diff
@@ -185,30 +184,35 @@ handle_cast({?RECONCILE_START, Options}, #state{conn_backoff = Backoff0} = State
             {noreply, State#state{conn_backoff = Backoff2}}
     end;
 handle_cast(
-    {?RECONCILE_END, #{forward_pid := Pid, commit := false} = Options, EUIPairs},
+    {?RECONCILE_END, #{error := Error} = Options, _EUIPairs},
```
michaeldjeffrey, line 187: I restructured the `reconcile_end` callbacks.

The list handler module will add an `error` key to `Options` if it gets a trailer other than `<<"0">>`, then the stream will get an `eos` and forward back to this module.

First we handle Errors:
- forward to the caller (if there's a pid)
- maybe schedule a reconnect (if there is not a pid)

Then non-commits:
- compare to local db of records
- forward to caller (if you can get here without a forward pid, congrats you made some logs)

And lastly, commits:
- compare to local db of records
- forward to caller
- maybe reschedule if the update fails (only if there's no pid)

src/grpc/router_ics_skf_worker.erl
```diff
@@ -131,67 +175,59 @@ handle_cast({?RECONCILE_START, Options}, #state{conn_backoff = Backoff0} = State
...
-            lager:info("reconciling done adding ~w removing ~w", [
-                erlang:length(ToAdd), erlang:length(ToRemove)
-            ]),
-            {noreply, State#state{reconciling = false}}
+            lager:info("reconciling done added ~w removed ~w", [AddCount, RemCount]),
+            {noreply, done_reconciling(State)}
     end;
 handle_cast(
-    {?UPDATE, _Updates}, #state{reconciling = true} = State
+    {?UPDATE, Updates}, #state{reconciling = true, deferred_updates = DeferredUpdates} = State
```
michaeldjeffrey, line 227: This module got the same `reconcile_end` restructuring as the EUI worker.

If an update comes through while we're reconciling, it will get queued. When reconciling is done, we'll process those updates. 
This prevents another entire reconcile that can take a long time and build up more queued updates, or dropping session keys on the floor to never be picked up.

config/sys.config.src
macpie: Can we move this under `{ics, #{`

macpie (dismissed):

rebar.config
michaeldjeffrey: Will discuss next week.

macpie (approved):

macpie:
Not needed anymore

PR #953, opened 2023-03-24
Fix helium_packet_service when getting an empty send_response

macpie:


src/grpc/helium_packet_service.erl
```diff
@@ -22,13 +22,19 @@ init(_Rpc, Stream) ->
 route(#envelope_up_v1_pb{data = {packet, PacketUp}}, StreamState) ->
     case verify(PacketUp) of
         false ->
+            lager:debug("failed to verify ~p", [PacketUp]),
             {grpc_error, {grpcbox_stream:code_to_status(2), <<"bad signature">>}};
```
michaeldjeffrey, line 26: Do we want to change this bad signature to a 7, so if this happens again we know it's not here.
```
code_to_status(7) -> ?GRPC_STATUS_PERMISSION_DENIED;
```

michaeldjeffrey (approved):

michaeldjeffrey (approved):

PR #959, opened 2023-04-11
Router unchained

michaeldjeffrey:


src/device/router_device_routing.erl
```diff
@@ -756,36 +756,31 @@ validate_devaddr(DevNum) when erlang:is_integer(DevNum) ->
...
+                        false -> {error, ?DEVADDR_NOT_IN_SUBNET}
+                    end
             end;
         _ ->
             {error, ?DEVADDR_MALFORMED}
     end.
 
+-spec get_subnets_bases() -> list({non_neg_integer(), non_neg_integer()}).
+get_subnets_bases() ->
+    _ = e2qc:cache(
```
macpie, line 774: can we try to go with `cream` here jus tin case?

config/test.config
```diff
@@ -14,7 +14,12 @@
         {sc_sup_type, server}
     ]},
     {grpcbox, [
-        {client, #{channels => [{default_channel, [{http, "localhost", 8080, []}], #{}}]}},
+        {client, #{
```
macpie, line 17: these probably also need to be added in `sys.config.src` ?

macpie (approved):

PR #982, opened 2023-07-26
base64 downlink payload in console events

michaeldjeffrey:


macpie:
(This will need some comms)

macpie (approved):

PR #1000, opened 2023-08-18
Free joins

macpie:


michaeldjeffrey (approved):

src/device/router_device_worker.erl
```diff
@@ -1278,12 +1278,27 @@ validate_join(
...
+                                Region,
+                                APIDevice,
+                                AppKey,
+                                Device
+                            ),
+                            {ok, UpdatedDevice, DevNonce, JoinAcceptArgs, {Balance, Nonce}}
+                    end;
+                false ->
+                    OrgID = maps:get(organization_id, router_device:metadata(Device), undefined),
+                    {Balance, Nonce} = router_console_dc_tracker:current_balance(OrgID),
```
macpie, line 1301: We need to heck that returning current balance won't affect Console

michaeldjeffrey: https://github.com/helium/console/blob/3763e1b38f69b099955526a28cb80762e5113e96/lib/console/etl_error_worker.ex#L96C19-L114

balance is only updated if `used` has a value `> 0`

michaeldjeffrey: If anything, it will be a DB call that sets the balance to what it already is.
There's a change in https://github.com/helium/console/pull/1279 to only update the balance if DC was used.

PR #1003, opened 2023-08-29
manage fcnt

michaeldjeffrey:


src/device/router_device_routing.erl
```diff
@@ -1185,80 +1182,66 @@ get_and_sort_devices(DevAddr, PubKeyBin) ->
 get_device_by_mic(_MIC, _Payload, []) ->
     undefined;
 get_device_by_mic(ExpectedMIC, Payload, [Device | Devices]) ->
-    ExpectedFCnt = expected_fcnt(Device, Payload),
```
michaeldjeffrey, line 1188: Don't guess at what the fcnt should be. 
Iterate through the higher frame bits and determine for certain what the frame count is. 
Then we can pass that fcnt through to the device worker and know it's verified.

src/device/router_device_routing.erl
```diff
@@ -1102,13 +1107,13 @@ send_to_device_worker_(FCnt, Packet, PacketTime, HoldTime, Pid, PubKeyBin, Regio
             end
     end.
 
--spec find_device(
+-spec find_device_for_data(
```
macpie, line 1110:
```suggestion
-spec find_device_for_uplink(
```

src/device/router_device_worker.erl
```diff
@@ -749,9 +747,7 @@ handle_cast(
     Device1 =
         case LastDevNonce == undefined of
             true ->
-                D1 = router_device:update([{fcnt, PacketFCnt}], Device0),
```
macpie, line 752: why remove this?

macpie (approved):

PR #1009, opened 2023-10-05
ets table per devaddr

michaeldjeffrey:
inserting devices into a ets bad where the whole object needs to be unique slows down severely as the table grows but the keys do not grow with the table. Inserting 250k devices spread across 32 devaddrs was approaching 5 minutes.

making a table per devaddr introduces a bit of indirection, but ets sets are fast, and device ids are unique.

### DIFF
--- a/src/device/router_device_cache.erl
+++ b/src/device/router_device_cache.erl
@@ -119,7 +119,7 @@ make_devaddr_table(DevAddr) ->
         {keypos, 2},
         {heir, whereis(router_db), DevAddr}
     ]),
-    ets:insert(?DEVADDR_ETS, {DevAddr, Ref}),
+    true = ets:insert(?DEVADDR_ETS, {DevAddr, Ref}),
     Ref.
 
 -spec init_from_db() -> ok.
--- a/src/device/router_device_cache.erl
+++ b/src/device/router_device_cache.erl
@@ -40,7 +40,7 @@ init() ->
     _ = ets:new(?DEVADDR_ETS, [
         public,
         named_table,
-        bag,
+        set,
         {write_concurrency, true},
         {read_concurrency, true}
     ]),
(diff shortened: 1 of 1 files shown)


michaeldjeffrey:
250k devices across 8 devaddrs takes ~8s to rehydrate locally.

macpie:
Should we only insert `device_id` in there? I would save memory and maybe avoid out of sync?

src/device/router_device_cache.erl
```diff
@@ -116,6 +109,19 @@ size() ->
...
 
+-spec make_devaddr_table(binary()) -> ets:tab().
+make_devaddr_table(DevAddr) ->
+    Ref = ets:new(devaddr_table, [
+        public,
+        set,
+        %% items are router_device:device()
+        %% keypos is the id field of the record.
+        {keypos, 2},
+        {heir, whereis(router_db), DevAddr}
```
macpie, line 120:
```suggestion
        {heir, whereis(router_sup), DevAddr}
```

src/device/router_device_cache.erl
```diff
@@ -116,6 +109,19 @@ size() ->
...
+    ]),
+    ets:insert(?DEVADDR_ETS, {DevAddr, Ref}),
```
macpie, line 122:
```suggestion
    true = ets:insert(?DEVADDR_ETS, {DevAddr, Ref}),
```

src/device/router_device_cache.erl
```diff
@@ -62,43 +62,36 @@ get(DeviceID) ->
...
+    ets:tab2list(?DEVADDR_ETS).
+
 -spec save(router_device:device()) -> {ok, router_device:device()}.
 save(Device) ->
     DeviceID = router_device:id(Device),
     true = ets:insert(?ETS, {DeviceID, Device}),
-    _ = erlang:spawn(fun() ->
-        % MS = ets:fun2ms(fun({_, D}=O) when D#device_v7.id == DeviceID -> O end),
-        MS = [{{'_', '$1'}, [{'==', {element, 2, '$1'}, {const, DeviceID}}], ['$_']}],
+    _ = erlang:spawn_link(fun() ->
```
macpie, line 75:
```suggestion
    _ = erlang:spawn_monitor(fun() ->
```

macpie (approved):

PR #1010, opened 2023-10-05
always look up the latest device in the main cache table

michaeldjeffrey:


macpie:
Does not apply anymore

